PR #6, opened 2017-08-18
Fetch and sync tracks when a webhook comes in

kntsoriano:
Hello, everyone! I would want to have your thoughts on my strategy to implement track updates via GitHub webhooks.

## Problem
We fetch and sync tracks at a constant interval. We want to change this to only fetch and sync when a track updates.

## Solution
Instead of syncing and fetching at every interval, we can use a GitHub webhook to signal that a track has updated.

In detail, here are the steps:

1. Create a push event webhook in git repos of our tracks.
2. Webserver receives webhook.
3. Webserver checks if webhook is a push to master.
4. If yes, webserver enqueues a background job that first fetches then syncs the tracks.
4. If no, do nothing.

## Incidental changes
1. Add push event webhooks to existing repos.

## Discussion
1. This change implies that `Git::SyncsTracks` doesn't run on `processor` anymore, leading to more load on our web servers. We may want to use [active-elastic-job](https://github.com/tawan/active-elastic-job) to spawn new instances whenever we need to sync tracks.
2. This code can be further optimized to fetch and sync only the specific track that has been updated. I decided to not to do it right now to quickly get feedback.
3. This code also lacks authentication.


kytrinyx:
This is a great plan. Checking that the push is to master is key. I agree with @iHiD that updating only the repo that got pushed to is need-to-have.

> This code also lacks authentication.

Hm. I don't think that's important, but I might not be aware of the risks. @kntsoriano would you list them out for me?

kytrinyx:
>  kytrinyx will be able to automatically add the push event webhooks to all existing repos for you when you're ready.

Yepp. Here's the script:

```
require 'octokit'

client = Octokit::Client.new(access_token: ENV['ADMIN_ORG_TOKEN'])
Octokit.auto_paginate = true

hook_url = "http://example.com/hook" # TODO
client.repos('exercism', :accept => 'application/vnd.github.mercy-preview+json', per_page: 100).each do |repo|
  next unless repo.topics.include?("exercism-track")

  client.create_hook(repo.full_name, 'web', {:url => hook_url, :content_type => 'json'}, {:events => ['push'], :active => true})
end
```

kytrinyx:
Another thing to consider as part of this is which caches will need to be invalidated after updating. Maintainer bios come to mind, but they are likely not the only bit of cached content.

kntsoriano:
Thank you for your thoughts.

Thinking about it again, the only risk I see of using no authentication is being spammed with fake requests to fetch repositories. It might not be a big concern after all.

@kytrinyx Would you be able to expound on caches that need to be invalidated? Where does this come in the code? Thank you very much :)

kytrinyx:
> To match which track to fetch and sync, I used the criteria that github repository name == track slug.

That's correct. This is one of our requirements, and should always remain true.

> would you be able to expound on caches that need to be invalidated?

I'm not actually sure what we're caching. My assumption is:

- maintainer bios
- track configuration
- track documentation
- track exercises

But I think that would need to be a question to @ccare, because I haven't looked at that part of the code at all.

>  I've seen in the code that every time a track is fetched, the website copy and problem specs are fetched, too.

Interesting, that shouldn't be necessary. We should probably have a hook that listens explicitly for changes to each of those repositories and update them when it happens.

I'm not even sure the site needs `problem-specifications`. We used to use the problem specification to generate exercise READMEs for all of the tracks on the fly, but we're no longer doing that.

ccare:
I need to read this all carefully, but there are a few things that jump out to me. We're going to have to discuss this on slack.

Firstly, we really want to run the sync code in one place predictably, which I think is sorted by the idea of it being a background job. So 👍 there. We do want this to run away from the webservers because it involves database sync and it would be good to have some degree of control over race conditions to that.

Secondly, we also (separately) need to orchestrate a git fetch (different from the sync) on each front-end webserver, not just the one that picks up the webhook. So we need to work that out too.

Thirdly: I've pushed to the master copy of FetchesTracks, sync code underneath you, sorry! But I will help with the merge.

Finally: I would like some failsafe code that does a fetch and sync every hour or so (instead of all the time) so we have a fallback should something ever go wrong with the integration. That would protect against the integration not being added to a track (or accidentally removed). I'm sure @kytrinyx can put tooling in to protect us from this, but we probably should defend out end too. 

In addition to the last point, we also need to consider what this looks like when we deploy to a clean server, that's another example of where we'll want to fetch files for all tracks.

Let's discuss it on slack when you have time :)

ccare:
```
I'm not even sure the site needs problem-specifications. We used to use the problem specification to generate exercise READMEs for all of the tracks on the fly, but we're no longer doing that.
```

^ you're right, we don't need this on the front-end servers anymore, but we are using it for syncing, to have a standardised description of a problem where it's not provided in the exercise metadata. Where there's no `Description.md` in folders like this: https://github.com/exercism/java/tree/master/exercises/hello-world/.meta, we fallback to the one in problem-specifications (e.g. https://github.com/exercism/problem-specifications/blob/master/exercises/hello-world/description.md)

kntsoriano:
@ccare Thanks :) Really good point about the failsafe code. Would be glad to discuss on Slack.

kytrinyx:
> we are using it for syncing, to have a standardised description of a problem where it's not provided in the exercise metadata. 

Ah, yepp. Forgot about that. Yes, we are using it for that!

kntsoriano:
Hello, everyone. I'm closing this PR in favor of this: https://github.com/exercism/prototype/pull/11

PR #9, opened 2017-08-20
Enable Bugsnag only on production

kntsoriano:
When `config.notify_release_environments` is not set, Bugsnag notifies
us of errors occurring in all the Rails environments. This is
particularly problematic when doing TDD as our bug monitoring will get
filled with false positives.

Here's an example of my errors appearing in the `bugsnag` channel:

screen shot 2017-08-20 at 3 31 41 pm


iHiD:
That's great. Thanks :)

PR #10, opened 2017-08-20
Fix failing handle regexp

kntsoriano:
When running the test suite on master, I noticed that
`test/models/user_test.rb:82` fails. I updated the regexp in order to
match the expectations described in the test.

Furthermore, I think failing tests on master suggests that we should add
a CI check before merging. What are your thoughts?


iHiD:
@kntsoriano Good catch. Sorry. The regexp was correct. The test was wrong. I simplified it based on a better understanding of Github's rules.

Definitely CI etc as soon as this is beyond prototype phase 💙

PR #11, opened 2017-08-24
Fetch and sync track updates when a webhook comes in (V2)

kntsoriano:
## Description
This is a reimplementation of fetching and syncing track updates. This was done because the first PR didn't take into account of the following:

1. Track update fetches need to happen on all webservers. The previous PR only does it on the webserver that receives the webhook.
2. Syncing should only happen when all webservers have fetched the track update. The previous PR has no way to guarantee this.

## Solution
Taking the above into account, here is a summary of what this implements:

1. When a webhook comes in, a `TrackUpdate` model is created. The `TrackUpdate` model is only created when it is a push to master.
2. The webservers poll for a fetchable `TrackUpdate`. A `TrackUpdate` is fetchable if it has no associated `TrackUpdateFetch` records that matches the webserver's host name.
3. When the webserver fetches a `TrackUpdate`, a `TrackUpdateFetch` record is created recording the webserver's host name. After fetching, the same `TrackUpdateFetch` record is marked with the time it has been fetched.
4. The processor polls for a syncable `TrackUpdate`. A `TrackUpdate` is syncable if the `TrackUpdate` has fetched `TrackUpdateFetches` equal to the number of webservers. This is currently set as `ENV[NUM_WEBSERVERS]` on the processor.
5. After the processor syncs a `TrackUpdate`, the `TrackUpdate` record is marked with the time it has been synced.

## Discussion
1. Since we set `ENV[NUM_WEBSERVERS]` on the processor, we should find a way to update this when we add a new webserver. I think we can do this later.

### DIFF
--- a/app/controllers/webhooks/track_updates_controller.rb
+++ b/app/controllers/webhooks/track_updates_controller.rb
@@ -1,0 +1,19 @@
+class Webhooks::TrackUpdatesController < WebhooksController
+  MASTER_REF = "refs/heads/master"
+
+  def create
+    TrackUpdate.create(track: track) if pushed_to_master?
+
+    render json: {}, status: 200
+  end
+
+  private
+
+  def pushed_to_master?
+    params[:ref].eql? MASTER_REF
+  end
+
+  def track
+    Track.find_by!(slug: params[:repository][:name])
+  end
+end
--- a/app/jobs/fetch_track_update_job.rb
+++ b/app/jobs/fetch_track_update_job.rb
@@ -1,0 +1,26 @@
+class FetchTrackUpdateJob < ApplicationJob
+  attr_accessor :track_update_fetch_id
+
+  before_perform do |job|
+    track_update = TrackUpdate.find(job.arguments.first)
+
+    track_update_fetch = TrackUpdateFetch.create!(
+      track_update: track_update,
+      host: Socket.gethostname
+    )
+
+    job.track_update_fetch_id = track_update_fetch.id
+  end
+
+  after_perform do |job|
+    track_update_fetch = TrackUpdateFetch.find(job.track_update_fetch_id)
+
+    track_update_fetch.update!(completed_at: Time.current)
+  end
+
+  def perform(track_update_id)
+    track_update = TrackUpdate.find(track_update_id)
+
+    Git::FetchesTracks.fetch([track_update.track])
+  end
+end
--- a/app/jobs/sync_track_update_job.rb
+++ b/app/jobs/sync_track_update_job.rb
@@ -1,0 +1,13 @@
+class SyncTrackUpdateJob < ApplicationJob
+  after_perform do |job|
+    track_update = TrackUpdate.find(job.arguments.first)
+
+    track_update.update!(synced_at: Time.current)
+  end
+
+  def perform(track_update_id)
+    track_update = TrackUpdate.find(track_update_id)
+
+    Git::SyncsTracks.sync([track_update.track])
+  end
+end
--- a/app/services/git/fetches_tracks.rb
+++ b/app/services/git/fetches_tracks.rb
@@ -1,6 +1,7 @@
 class Git::FetchesTracks
 
   ERROR_BACKOFF_PERIOD = 10.seconds
+  TRACK_BACKOFF_PERIOD = 1.second
 
   def self.fetch(tracks)
     new(tracks).fetch
@@ -15,7 +16,7 @@ class Git::FetchesTracks
     fetch_website_content
     tracks.each do |track|
       fetch_track(track)
-      sleep 1.second
+      sleep TRACK_BACKOFF_PERIOD
     end
   end
 
--- a/app/services/git/fetches_tracks.rb
+++ b/app/services/git/fetches_tracks.rb
@@ -1,25 +1,26 @@
 class Git::FetchesTracks
 
-  QUIET_PERIOD = 30.seconds
   ERROR_BACKOFF_PERIOD = 10.seconds
 
-  def self.run
-    new.run
+  def self.fetch(tracks)
+    new(tracks).fetch
   end
 
-  def run
-    loop do
-      fetch_problem_specs
-      fetch_website_content
-      Track.find_each do |track|
-        fetch(track)
-        sleep 1.second
-      end
-      sleep QUIET_PERIOD
+  def initialize(tracks)
+    @tracks = tracks
+  end
+
+  def fetch
+    fetch_problem_specs
+    fetch_website_content
+    tracks.each do |track|
+      fetch_track(track)
+      sleep 1.second
     end
   end
 
   private
+  attr_reader :tracks
 
   def fetch_problem_specs
     Rails.logger.info "Fetching problem specs"
@@ -41,7 +42,7 @@ class Git::FetchesTracks
     sleep ERROR_BACKOFF_PERIOD
   end
 
-  def fetch(track)
+  def fetch_track(track)
     repo_url = track.repo_url
     if repo_url.start_with?("http://example.com")
       Rails.logger.info "Skipping fetch for #{repo_url}"
--- a/app/services/git/fetches_updated_tracks.rb
+++ b/app/services/git/fetches_updated_tracks.rb
@@ -1,0 +1,44 @@
+class Git::FetchesUpdatedTracks
+
+  QUIET_PERIOD = 30.seconds
+
+  def self.run
+    new.run
+  end
+
+  def self.fetch
+    new.fetch
+  end
+
+  def run
+    loop do
+      fetch
+      sleep QUIET_PERIOD
+    end
+  end
+
+  def fetch
+    find_track_update
+    fetch_track_update if track_update
+  end
+
+  private
+
+  attr_reader :track_update, :track_update_fetch
+
+  def find_track_update
+   @track_update = TrackUpdate.
+     includes(:track_update_fetches).
+     where(synced_at: nil).
+     find_by(track_update_fetches: { track_update_id: nil }) ||
+     TrackUpdate.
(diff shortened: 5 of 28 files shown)


ccare:
I love the way you communicate on your PRs @kntsoriano, so clear. Thank you! 

Regarding, the number of webservers, I wonder if we should do this as a yaml file, rather than an env var. This isn't really a strong opinion, but based on two things:

1) By using a config file we can ensure it's set for every rails engine we spawn (webserver, rake task, sidekiq processor). I get that it's probably not needed everywhere, but it feels prudent to ensure that.

2) We then have only one approach to managing external config on exercism

Don't get me wrong, I love ENV vars, and they totally make sense, so this can be done if you feel strongly. What do you think?

app/services/git/syncs_updated_tracks.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def self.sync
+    new.sync
+  end
+
+  def run
+    loop do
+      sync
+      puts "Sleeping"
```
ccare: I need to get the rails logger working properly for rake tasks. It's a good idea to stick with PUTS for now, but I'm hoping to have that sorted today.

app/jobs/sync_track_update_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  after_perform do |job|
+    track_update = TrackUpdate.find(job.arguments.first)
+
+    track_update.update!(synced_at: Time.current)
+  end
+
+  def perform(track_update_id)
+    track_update = TrackUpdate.find(track_update_id)
+
+    Git::SyncsTracks.sync([track_update.track])
```
ccare: I wonder if we can just remove the `SyncsTracks` class now that we're not looping through everything? What do you think. There's obviously that hacky bit of code about `hello-world` to address, but that can go inside `SyncsTrack` I think.

app/services/git/syncs_updated_tracks.rb
```diff
@@ -0,0 +1,44 @@
...
+  private
+  attr_reader :track_update
+
+  def find_track_update
+    @track_update = TrackUpdate.
+      joins(:track_update_fetches).
+      where(synced_at: nil).
+      where.not(track_update_fetches: { completed_at: nil }).
+      group("track_update_fetches.id").
+      having("count(track_update_id) = #{ENV["NUM_WEBSERVERS"]}").
```
ccare: Nice use of `having` there, I still find active record queries hard to do. As I mentioned in the top-level comment. I wonder if we should externalise this in a YAML file so it can be set on a per-env basis. That's the pattern we've been following throughout this project, but happy to be overruled if the env var has advantages. See https://github.com/exercism/prototype/blob/master/app/models/git/github_profile.rb#L4 and https://github.com/exercism/prototype/blob/master/config/auth.yml for an example. It might be we want to refactor some of these configs and have one for managing the git integration stuff, maybe...

app/services/git/fetches_tracks.rb
```diff
@@ -1,25 +1,26 @@
...
-        sleep 1.second
-      end
-      sleep QUIET_PERIOD
+  def initialize(tracks)
+    @tracks = tracks
+  end
+
+  def fetch
+    fetch_problem_specs
+    fetch_website_content
```
ccare: We need to address this two fetches, because they're not tracks, but are specialisations of Git::RepoBase and therefore need to be fetched and triggered by a webhook - sorry for the scope creep!

app/services/git/fetches_updated_tracks.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  private
+
+  attr_reader :track_update, :track_update_fetch
+
+  def find_track_update
+   @track_update = TrackUpdate.
+     includes(:track_update_fetches).
+     where(synced_at: nil).
+     find_by(track_update_fetches: { track_update_id: nil }) ||
```
ccare: I find your code very readable @kntsoriano. This method did confuse me a bit though, wonder if the `||` here could be made a little clearer, I didn't spot it at first.

app/services/git/syncs_tracks.rb
```diff
@@ -1,41 +1,24 @@
 class Git::SyncsTracks
```
ccare: I think this class is almost redundant now. Same for `FetchesTracks` possibly, and the `Git::StateDb`. Your solution is much better.

app/jobs/fetch_track_update_job.rb
```diff
@@ -0,0 +1,26 @@
+class FetchTrackUpdateJob < ApplicationJob
+  attr_accessor :track_update_fetch_id
+
+  before_perform do |job|
+    track_update = TrackUpdate.find(job.arguments.first)
+
+    track_update_fetch = TrackUpdateFetch.create!(
+      track_update: track_update,
+      host: Socket.gethostname
```
ccare: We need to think about Socket.gethostname in an ec2 context, but let's defer that until we've sorted everything else out. There are a couple of options here which would be fun to discuss separately.

app/services/git/fetches_tracks.rb
```diff
@@ -1,25 +1,26 @@
...
+    tracks.each do |track|
+      fetch_track(track)
+      sleep 1.second
```
ccare: This my code not yours, but I should probably have extracted the backoff into a constant.

app/controllers/webhooks/track_updates_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  def create
+    TrackUpdate.create(track: track) if pushed_to_master?
+
+    render json: {}, status: 200
+  end
+
+  private
+
+  def pushed_to_master?
+    params[:ref].eql? MASTER_REF
```
ccare: 👍 I suspect I would have missed this subtlety

app/services/git/fetches_tracks.rb
ccare: Maybe `TrackUpdateFetches` should be generalised to `GitUpdateFetches` ?

app/jobs/sync_track_update_job.rb
kntsoriano: In the previous PR, you mentioned a nightly task that pulls tracks just in case the webhooks don't work. I think having the `SyncsTracks` service helps there since we can only call `SyncsTracks.sync(Track.all)`.

app/services/git/syncs_updated_tracks.rb
kntsoriano: Sure, I don't know any advantages of having an ENV var over a YAML file. I'll be glad to change this to use the YAML file.

app/services/git/fetches_tracks.rb
kntsoriano: I think we can stick with this code now and just generalize to `GitUpdateFetches` :)

app/services/git/fetches_tracks.rb
kntsoriano: No problem! I'll extract that to a constant.

app/services/git/fetches_updated_tracks.rb
kntsoriano: Sure, I'll come up with a way.

app/services/git/syncs_tracks.rb
kntsoriano: I think we should still keep this in order to have an easy way to sync all tracks, just as I have described for `FetchesTracks`. What do you think?

kntsoriano:
I have no opinions of using an ENV var over a YAML file. Since YAML files are what we use right now, I'll update my code to use it. Thanks for that :)

app/jobs/fetch_track_update_job.rb
kntsoriano: I think we can talk about this now. I've addressed your comments and I've got it working locally. This is the last bit we need to work on before we can deploy.

app/services/git/fetches_tracks.rb
ccare: I don't think so - what if just the website content changes? That should trigger a webhook which triggers a refresh of that content.

app/jobs/sync_track_update_job.rb
ccare: Good point. I see

app/jobs/fetch_track_update_job.rb
ccare: so there's no harm using the hostname, but it'll change on reboot (unless we pin internal addresses). Alternatively, we could lookup the public IP from the EC2 api (static and straightforward), or we could get chef to write a file somewhere that configures a server slug.

ccare: Would it matter if the hostname changed on reboot? I'm assuming so, but maybe not.

kntsoriano: I think it matters because we need to know whether the host has fetched the server or not. Does the public IP change only when we use new instance? It could also be a problem when the machine's IP changes while the instance doesn't.

app/services/git/fetches_tracks.rb
kntsoriano: Good point. I'd need to write some code to account for the problem specs and website copy repos changing.

app/jobs/fetch_track_update_job.rb
ccare: We do have fairly static IPs, but lets decouple this from the network identity, because as you hint, it's problematic. I'll adjust chef to write a slug into `/etc/exercism/server_identity` so if you just read from that file (and chomp) then we're good I think.

kntsoriano:
Hello, I figured that renaming tables and dropping foreign keys would be too complicated to continue on this branch. Please refer to the new pull request I opened: https://github.com/exercism/prototype/pull/16

PR #14, opened 2017-08-27
Add purescript

kytrinyx:
This track went live recently in production.


iHiD:
I'm going to hold off this until we work out what to do on the front page with the honeycomb else it'll break it and look awful. I'll sort that in the next day or two.

kytrinyx:
I'm going to close this and submit PRs to address both the active and inactive tracks.

PR #15, opened 2017-08-28
Remove hyphens from solution uuids

kytrinyx:
Per our discussion about search engines and SEO.

I added a comment to the code, since this is context that cannot be understood by reading the code itself.


iHiD:
Great. Thanks. Merge away

PR #20, opened 2017-09-06
Update repos at a regular interval

kntsoriano:
Create `RepoUpdate` records to update tracks, problem specifications, and
website copy. `rake git:update_repos` will be ran at a regular interval via a cron job.


ccare:
Thanks @kntsoriano, this looks great. If you merge it down and deploy it, we can sort out the cron job trigger.

PR #24, opened 2017-09-22
Add filters to mentor dashboard

kntsoriano:
To close https://github.com/exercism/reboot/issues/138, this PR adds Track and Exercise filters to the mentor dashboard.

## Solution
1. In order to allow for more filters to be added, refactor and rename `RetrievesSolutionsForMentor` to `FilterSolutionsByStatus`.
2. Add `RetrievesSolutionsForMentor` to accept additional options to filter by status, track_id, and exercise_id.
3. When a track is chosen from the filter, fetch the exercises associated from the track.
4. Run a JS script to update the options of the exercise select box based on the fetched exercises.

## Screencast
sep-22-2017 15-13-11

## Discussion
1. In order to preserve the old behavior, I decided to write a system test in order to be confident that I don't break anything.
2. Since we are using a JS based select plugin, I needed to write `SelectizeHelpers` in order to do actions on the select box.
3. `OptionsHelper` was added in order to add convenience method to serialize collections into a format understandable by the `Selectize` plugin, as well as Rails' `option_for_select`.

### DIFF
--- a/app/assets/javascripts/options.js
+++ b/app/assets/javascripts/options.js
@@ -1,0 +1,4 @@
+window.updateOptions = function(select, options) {
+  select.clearOptions();
+  select.addOption(options);
+}
--- a/app/assets/stylesheets/pages/mentor-dashboard.scss
+++ b/app/assets/stylesheets/pages/mentor-dashboard.scss
@@ -40,7 +40,6 @@
         margin-bottom:40px;
     }
     .widget-filter {
-        float:right;
         margin-right:0;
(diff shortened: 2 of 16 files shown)


kntsoriano:
Good suggestion :) We can change it once we deploy, shouldn't be blocking launch anyway. We'd need to add some way to do sizing for our filters once we put that in. It may be by adding modifier classes (`-m`, `-xl`, etc), or wrapping all our filters within pure grids.

PR #28, opened 2017-09-23
Implement webhook listener to track contributors

kytrinyx:
**NOTE - this relies on the migration in https://github.com/exercism/prototype/pull/31 to be merged, deployed, and run.** This PR is built on top of the migration in question, but if we deploy before migrating, then we risk getting some 500 errors.

### Background

In the footer of the site we show a count of contributors. We need this number to be incremented
when someone new contributes to a repository in the Exercism organization.

When we run the ETL for v2, we will need to run the existing contributor script _once_ to backfill existing data, and then we can rely on the webhook to keep us up-to-date.

### Implementation

I ran into some interesting snags in this one.

* `params` vs `request.request_parameters`
* passing an actual post body in a Rails controller test
* `request.request_parameters` vs `request.raw_post`

#### Params vs request parameters

The JSON POST body is available via params, however everything is stringified. If you passed `{"ok": true, "number": 1}`, then the params hash has `{"ok": "true", "number": "1"}`.

The fix is to access the JSON POST body via `request.request_parameters`, which retains the original types for the values.

#### Testing with a POST body

The entire internet says to pass `params`:

```
post webhooks_contributors_path, params: {"ok": true, "number": 1}
```

That's fine if you're ok with stringified JSON values. Not so great if we need the original.

I dug into the Rails source code and figured out that the POST body can be passed as a key within the environment: `action_dispatch.request.request_parameters`. This should be the hash version, not the JSON version.

But this brings me to the trickiest bit.

#### Request parameters vs raw post

We need to verify the signature of the request. That means that we need to combine the original JSON string with our webhook secret, and do some hashing, and then compare the result to the signature that we get in the header. Saying `request.request_parameters.to_json` doesn't cut it. There are small differences in formatting, so we end up with a different signature.

Back to digging in the source, it turns out the request provides access to the original POST body: `request.raw_post`. We can use this value to calculate the signature.

### DIFF
--- a/app/controllers/webhooks/contributors_controller.rb
+++ b/app/controllers/webhooks/contributors_controller.rb
@@ -1,0 +1,44 @@
+require_relative '../../../lib/open_source'
+
+class Webhooks::ContributorsController < WebhooksController
+  before_action :verify_github_webhook
+
+  def create
+    contribution = OpenSource::Contribution.new(request.request_parameters)
+
+    unless contribution.complete?
+      head :no_content and return
+    end
+
+    contributor = Contributor.with_contribution(contribution)
+
+    if contributor.save
+      head :no_content
+    else
+      message = "Failed to save event with X-GitHub-Delivery: %s" % github_delivery
+      render json: {error: message}, status: 500
+    end
+  end
+
+  private
+
+  def verify_github_webhook
+    payload = request.raw_post
+    signature = 'sha1=' + OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha1'), secret, payload)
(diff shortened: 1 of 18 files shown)


kytrinyx:
This is ready for review.

kytrinyx:
@kntsoriano I don't want to merge this until the related DB migration is approved, merged, deployed, and run, so I'll let @iHiD take this one when he has time (no hurry).

iHiD:
> Until the related DB migration is approved, merged, deployed, and run

Approved. Merged. Deployed. Run :) 

Would appreciate you responding to my comment still, just as it piqued my interest.

PR #29, opened 2017-09-23
Include the test server identity

kytrinyx:
We only want to ignore the server_identity file for development use.

The tests should have a fixture by default, otherwise we make new developers jump through
unnecessary hoops to get the tests running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,4 @@
 
 /app/assets/images/tmp
 
-server_identity
+/server_identity
--- a/test/fixtures/server_identity
+++ b/test/fixtures/server_identity
@@ -1,0 +1,1 @@
+hostname


iHiD:
So, I've just pulled and run the tests and hit the same thing. For reference, the commit that caused this is [here](https://github.com/exercism/prototype/commit/e71d039be1f62d756232760f411c64cfa095453d).

I'm a :+1: on this but would like @kntsoriano to review as he added this.

kytrinyx:
Awesome. I'm in no way convinced this is the right solution, but it was the first thing I thought of that seemed like it would work :)

iHiD:
@kntsoriano Your solution (README) worked for development, but not tests, I think. We'll still need the README comment for the development setup atm I believe. @ccare - Can you give this a final nod pls?

ccare:
Fine with me, so long as production continues to wire back to the path here https://github.com/exercism/prototype/blob/6bd01f5d3185e1af9a0b1ffac4246dbe848e2fce/config/cluster.yml#L7

kytrinyx:
@kntsoriano @iHiD Yepp—your solution was good for dev, but test needed some extra love.

I'll go ahead and merge this. Thanks all!

PR #30, opened 2017-09-23
[migrations] Update contributors table

kytrinyx:
This has two changes:

* Add a default value of 0 to the `num_contributions` column so that we don't have to jump through hoops to deal with NULL values.
* Add a GitHub ID, so that we can use that as a source of truth, since people can (and do) change their GitHub username.

This is a prerequisite for pull request #28, and should be merged and deployed before #28 lands.


iHiD:
Thanks. Merged. My local setup is screwed atm, so I can't deploy. But will do once I've worked that out.

kytrinyx:
Awesome, thanks! I'll have the rest of this feature in a PR by Monday morning your time.

iHiD:
Great. Thanks. Deployed :)

PR #31, opened 2017-09-23
[migrations] Don't allow nulls in GitHub ID for contributors

kytrinyx:
I missed a spot :/

This is for https://github.com/exercism/prototype/pull/28


iHiD:
@kytrinyx What's the need for the `require_relative` change, out of interest? (Not against it - just don't understand why it's needed :))

kytrinyx:
> What's the need for the require_relative change, out of interest? 

I run individual test files rather than the entire suite, as it's significantly faster. If I don't do this, then it says "file not found".

PR #32, opened 2017-09-25
Update schema.rb

kntsoriano:
As I did a fresh setup of my DB, I noticed that db/schema.rb is not
up to date on master.


iHiD:
Thanks.

PR #34, opened 2017-11-09
Add titles to external pages

kntsoriano:
As discussed in https://github.com/exercism/reboot/issues/109, this PR adds page titles to external pages for the purposes of SEO.

As I haven't had any experience with SEO yet, I decided to follow this [resource](https://moz.com/learn/seo/title-tag) to format our title tags properly.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -15,7 +15,7 @@ class PagesController < ApplicationController
     define_method page do
       markdown = Git::WebsiteContent.head.pages["#{page}.md"] || ""
       @page = page
-      @title = title
+      @page_title = title
       @content = ParsesMarkdown.parse(markdown.to_s)
       render action: "generic"
     end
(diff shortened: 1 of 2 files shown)


app/helpers/metadata_helper.rb
```diff
@@ -27,19 +34,34 @@ def metadata
...
-            }
+            { title: @title }
           end
         when "tracks"
           case action_name
+          when "index"
+            { title: "Language Tracks" }
           when "show"
             {
-              title: "#{@track.title} on Exercism",
```
iHiD: What's the logic behind taking this out?

app/controllers/pages_controller.rb
```diff
@@ -27,6 +27,7 @@ def index
   end
 
   def team
+    @title = "The Exercism Team"
```
iHiD: I think we should name this variable `@page_title` or `@meta_title` or something, to avoid risking clashes. Thoughts?

iHiD: Also, what's your thinking about doing this here rather than in the helper like the other ones?

kntsoriano: I decided to put this here since all actions which goes through the page controller uses the `@title` variable.  I agree with changing the variable name to `@page_title`.

app/helpers/metadata_helper.rb
kntsoriano: This will be appended with `| Exercism` as followed in the SEO guidelines I posted above.

iHiD: Ah. I missed that method. Gotcha.

PR #35, opened 2017-11-09
Skip mentor rating

kntsoriano:
As discussed in https://github.com/exercism/reboot/issues/79, this hides the mentor rating pane when completing an auto approved solution.

Screencast: http://sendvid.com/xksfs0z7


iHiD:
`sendvid.com` seems quite spammy btw. Opens popups and stuff.

PR #39, opened 2017-11-16
Fix bug with Git::SeedsTracks calling the wrong class

kntsoriano:
When I was setting up a fresh repository, I noticed that `Git::SeedsTracks` was calling a non existent class called `Git::FetchTrack`.

### DIFF
--- a/app/services/git/fetches_repo.rb
+++ b/app/services/git/fetches_repo.rb
@@ -1,0 +1,25 @@
+class Git::FetchesRepo
+  ERROR_BACKOFF_PERIOD = 10.seconds
+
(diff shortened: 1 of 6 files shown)


ccare:
Looks fine to me, although presumably we only now seed during dev setup? Or is this production code.

kntsoriano:
@ccare Seeds are only used when setting up development :)

kytrinyx:
Would it be OK if we merge this? I need to seed my local database with tracks when running the ETL from v1 to v2.

kytrinyx:
@iHiD says go for it. So I'm going for it!

PR #43, opened 2017-12-25
Fix runtime error on db:seed

kytrinyx:
When processing a track's config, we were using

     exercise.fetch(:topics, [])

to get the topics. However, some tracks explicitly set the topics to null
if they've not defined the topics, which meant that fetch was succeeding
and not falling back to the default. This caused the map on the following
line to blow up.


iHiD:
I'm :+1: on this.

PR #50, opened 2018-01-01
Add support for inactive tracks

kytrinyx:
**depends on migration in https://github.com/exercism/prototype/pull/49** -- update: deployed/migrated

This stores the active flag on tracks based on their `config.json`, and supports an `active` scope on the `Track` model.

It scopes the Track collections to the active tracks on:

- the homepage
- the track list page
- the "my tracks" list page

This does not change any endpoints that act on a specific track, which means that you could go to the page for an inactive track, or fetch/submit solutions for inactive tracks, if you go to the correct URL or pass the right parameters.

Lastly, it adds all the currently inactive tracks to the list of tracks to be seeded with `rake db:seed`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,8 +50,9 @@ group :development do
 end
 
 group :test do
-  gem 'mocha'
   gem 'minitest', '~> 5.10', '!= 5.10.2'
-  gem 'timecop'
+  gem 'minitest-stub-const'
+  gem 'mocha'
   gem 'rails-controller-testing'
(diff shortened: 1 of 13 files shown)


kytrinyx:
> I'd like to scope that to maintainers (which we have a table for) but that's not a blocker for me at all. So let's get it merged down and I'll open an issue for us to do that at a later stage?

That sounds excellent. Go ahead and open the issue and I'll tackle it soon-ish.

PR #52, opened 2018-05-08
Upgrade loofah

kytrinyx:
This includes some important updates to the loofah gem. The newer
version of the gem required a dependent update of nokogiri.

I ran the tests, but didn't boot the app up and run through any
scenarios. If we are unsure about the test coverage for the features
using loofah, we may want to do a sanity check before merging.


kytrinyx:
Actually, belay this. We have another critical update we need, which will pull in these upgrades transitively.

PR #53, opened 2018-05-08
Upgrade rails-html-sanitizer

kytrinyx:
This updates dependent gems as well, of particular note: loofah and nokogiri.

I ran the tests, but didn't boot the app up and run through any
scenarios. If we are unsure about the test coverage for the features
using loofah, we may want to do a sanity check before merging.

Replaces https://github.com/exercism/prototype/pull/52


iHiD:
I'm just going to merge this and we'll test it along with everything else we'll be testing in the next few weeks.

PR #58, opened 2018-06-02
Read interactive walkthrough from new location

kytrinyx:
Per decision documented in https://github.com/exercism/reboot/issues/248

I went with `walkthrough/index.html` rather than `walkthrough.html` because (a) it feels a bit cleaner, and (b) we already have beautiful abstractions for reading from a folder, and I'm not sure if a folder reader would work with `.`.

### DIFF
--- a/app/controllers/my/solutions_controller.rb
+++ b/app/controllers/my/solutions_controller.rb
@@ -29,7 +29,7 @@ class My::SolutionsController < MyController
   end
 
   def walkthrough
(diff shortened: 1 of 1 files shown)


kytrinyx:
The build is failing. I'll take a look in a bit.

kytrinyx:
It's failing on master, too, it looks like.

kytrinyx:
Ok, got it. Pushed a fix to master, rebased this onto master. Should be good to go.

PR #60, opened 2018-06-13
Send email when creating a team invite

kntsoriano:
Closes https://github.com/exercism/reboot/issues/233.

## Description
This PR sends an email when a `TeamInvitation` is created via the `CreateTeamInvitation` service. As described in the issue linked, different email content is sent whether the invitee's email has been registered or not.

## Discussion
1. I think we need to discuss what the body of these emails should be. I've just done the bare minimum. If it's not that important right now, we can amend changes in subsequent PRs.

### DIFF
--- a/test/services/create_team_invitation_test.rb
+++ b/test/services/create_team_invitation_test.rb
@@ -1,0 +1,16 @@
+require 'test_helper'
+
+class CreateTeamInvitationTest < ActiveSupport::TestCase
+  test "creates team invitation" do
+    inviter = create(:user)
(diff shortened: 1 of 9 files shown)


test/services/create_team_invitation_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class CreatsTeamInvitationTest < ActiveSupport::TestCase
```
iHiD: /s/Creats/Create

PR #61, opened 2018-06-13
Join team via URL

kntsoriano:
Closes https://github.com/exercism/reboot/issues/235 and https://github.com/exercism/reboot/issues/236.

## Description
This PR does two things:

1. It allows team admins to get a URL to share with people they want to be part of their team.
2. It allows said people to join the team via the URL shared.

### Generating the join URL
jun-13-2018 17-42-23
### Joining a team
jun-13-2018 17-43-03


app/assets/stylesheets/section.scss
```diff
@@ -0,0 +1,7 @@
+.section {
```
iHiD: This is too broad a change as it will change every `.section` within the whole platform. I'd rather insert this into the `team-page.scss` file so the changes only affect that single page.

app/assets/stylesheets/section.scss
```diff
@@ -0,0 +1,7 @@
...
+  margin-bottom: 2rem;
```
iHiD: Could you use `px` for units pls. I'm open to being told that `rem` is better, but everything else is `px` atm so I'd like to keep it consistent until/unless we change it everywhere.

app/assets/stylesheets/teams/join.scss
```diff
@@ -0,0 +1,3 @@
+.team-join-page {
+  margin-top: 2rem;
```
iHiD: `px` again pls

app/controllers/teams/joins_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  def show
+    team = Team.find_by!(token: params[:team_token], url_join_allowed: true)
+
+    @team_membership = TeamMembership.new(team: team)
+  end
+
+  def create
+    team = Team.find_by!(token: params[:team_token], url_join_allowed: true)
+
+    TeamMembership.create!(team: team, user: current_user)
```
iHiD: Can you wrap this in an exception handler and redirect back to `:show` if it fails pls. (An example of why this might fail is if someone double clicks, so the first one creates the team and the second one then blows up, but that second response is the response the person gets. With this `rescue` and the guard above we avoid that scenario)

app/controllers/teams/joins_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Teams::JoinsController < TeamsController
+  skip_before_action :find_team
+
+  def show
+    team = Team.find_by!(token: params[:team_token], url_join_allowed: true)
```
iHiD: Can you add a guard here so if someone is already a member of this team, they get redirected to the team page please.

app/models/team.rb
```diff
@@ -15,6 +15,14 @@ def slug_candidates
...
   has_many :memberships, class_name: "TeamMembership"
   has_many :members, through: :memberships, source: :user
+  has_many :admins,
+    -> { where(team_memberships: { admin: true }) },
+    through: :memberships,
+    source: :user
 
   has_secure_token :token
+
+  def admin?(user)
```
iHiD: Could you rename this to `has_admin?` please. That parses better to me so in my head instead of saying  "Hello Team. Are you admin User:Karlo?" it instead says "Hello team, do you have User:Karlo as an admin?"

PR #62, opened 2018-06-14
Track fetching improvements

kntsoriano:
## Description
This PR contains fixes for fetching updates from tracks.

### Fix bug with fetching an already fetched repo update
We weren't able to sync tracks correctly because the `git:fetch` task crashed for one of the hosts. The reason why it crashed was that it was trying to fetch an already fetched update. In order to fix this, I had to rewrite the SQL query to look for tracks already fetched.

### Fetch all repo updates in a poll period
As we've needed to catch up with the updates we haven't fetched (due to the scenario described above), I noticed that the way we've written `git:fetch` was inefficient. It would take me a long time to get up to date with all the track updates! I've rewritten `git:fetch` to fetch all repo updates in a poll period, rather than fetching one update per poll period.

### DIFF
--- a/app/services/git/fetches_updated_repos.rb
+++ b/app/services/git/fetches_updated_repos.rb
@@ -18,31 +18,29 @@ class Git::FetchesUpdatedRepos
   end
 
   def fetch
-    find_repo_update
-    fetch_repo_update if repo_update
+    repo_updates.each { |update| FetchRepoUpdateJob.perform_now(update.id) }
   end
 
   private
 
   attr_reader :repo_update, :repo_update_fetch
 
-  def find_repo_update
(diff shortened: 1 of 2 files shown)


app/services/git/fetches_updated_repos.rb
```diff
@@ -18,31 +18,29 @@ def run
   end
 
   def fetch
-    find_repo_update
-    fetch_repo_update if repo_update
+    repo_updates.each { |update| FetchRepoUpdateJob.perform_now(update.id) }
```
ccare: Is there a potential need for de-duping fetches at some point? Say if there are N hook events for the same repo in rapid succession?

PR #63, opened 2018-06-14
Remove team member

kntsoriano:
Closes https://github.com/exercism/reboot/issues/238.

## Description
This PR implements team member removal. This action is only allowed for team admins.

## Screencast
jun-14-2018 19-37-57

### DIFF
--- a/app/controllers/teams/team_memberships_controller.rb
+++ b/app/controllers/teams/team_memberships_controller.rb
@@ -2,16 +2,12 @@ class Teams::TeamMembershipsController < ApplicationController
   before_action :authenticate_user!
 
   def destroy
-    team_membership = teams_managed_members.find(params[:id])
+    team_membership = TeamMembership.
+      where(team_id: current_user.teams_managed.pluck(:id)).
+      find(params[:id])
 
     team_membership.destroy
 
     redirect_to teams_team_memberships_path(team_membership.team)
-  end
-
-  private
-
-  def teams_managed_members
-    TeamMembership.where(team_id: current_user.teams_managed.pluck(:id))
   end
 end
--- a/app/controllers/teams/team_memberships_controller.rb
+++ b/app/controllers/teams/team_memberships_controller.rb
@@ -2,10 +2,16 @@ class Teams::TeamMembershipsController < ApplicationController
   before_action :authenticate_user!
 
   def destroy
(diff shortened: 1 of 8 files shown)


app/controllers/teams/team_memberships_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    team_membership.destroy
+
+    redirect_to teams_team_memberships_path(team_membership.team)
+  end
+
+  private
+
+  def teams_managed_members
+    TeamMembership.where(team_id: current_user.teams_managed.pluck(:id))
```
iHiD: `select` will be more efficient than `pluck` here, as it will turn into a nested query in AR.

app/controllers/teams/team_memberships_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    team_membership = teams_managed_members.find(params[:id])
+
+    team_membership.destroy
+
+    redirect_to teams_team_memberships_path(team_membership.team)
+  end
+
+  private
+
+  def teams_managed_members
```
iHiD: I don't think this needs to be it's own method. It just adds indirection. I think a single statement is more readable: 
```
TeamMembership.where(team_id: current_user.teams_managed.select(:id)).
               find(params[:id])
```

app/models/user.rb
```diff
@@ -19,6 +19,7 @@ def self.validates_uniqueness_of(*args)
 
   has_many :team_memberships
   has_many :teams, through: :team_memberships
+  has_many :teams_managed, -> { where(team_memberships: { admin: true }) }, through: :team_memberships, source: :team
```
iHiD: What do you think about renaming this to `managed_teams`. I think it's clearer.

app/controllers/teams/team_memberships_controller.rb
iHiD: Check the queries for
```
TeamMembership.where(team_id: current_user.teams_managed.pluck(:id))
```
vs
```
TeamMembership.where(team_id: current_user.teams_managed.select(:id))
```
They'll be different.
- `.pluck` returns an array
- `.select` returns an ActiveRecordRelation
In the second example above, AR will use that relationship to make a subquery. Whereas in the first example, it will pass ids in.

PR #65, opened 2018-06-16
Add validate_token API endpoint

iHiD:
To solve this issue in the CLI: exercism/cli#537

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,7 @@ Rails.application.routes.draw do
   namespace :api do
     scope :v1 do
       get "ping" => "ping#index"
+      get "validate_token" => "validate_token#index"
 
       resource :cli_settings, only: [:show]
       resources :tracks, only: [:show]
--- a/test/controllers/api/validate_token_test.rb
+++ b/test/controllers/api/validate_token_test.rb
@@ -1,0 +1,15 @@
+require_relative './test_base'
+
+class API::PingControllerTest < API::TestBase
+  test "setup should return 401 with incorrect token" do
+    get api_validate_token_path, as: :json
+    assert_response 401
+  end
+
+  test "latest should return 200 with correct token" do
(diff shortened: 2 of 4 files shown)


test/controllers/api/validate_token_test.rb
```diff
@@ -0,0 +1,15 @@
+require_relative './test_base'
+
+class API::PingControllerTest < API::TestBase
+  test "setup should return 401 with incorrect token" do
+    get api_validate_token_path, as: :json
+    assert_response 401
```
kytrinyx: I was originally thinking 403 but you're right. It needs to be a 401 because if your token is invalid we don't know who you are. Nice!

test/controllers/api/validate_token_test.rb
```diff
@@ -0,0 +1,15 @@
+require_relative './test_base'
+
+class API::PingControllerTest < API::TestBase
```
kytrinyx: D'oh. I didn't notice this but the class name should probably be `API::ValidateTokenControllerTest`

config/routes.rb
```diff
@@ -10,6 +10,7 @@
   namespace :api do
     scope :v1 do
       get "ping" => "ping#index"
+      get "validate_token" => "validate_token#index"
```
kytrinyx: I also didn't think to hard about this, but I'd like to consider whether we can make this more resource-y rather than RPC-y.

kytrinyx: Another thing: I prefer to normalize on URLs with hyphens not underscores, because:

- slug normalization uses hyphens, which means that when we parameterize things in the API, if we use underscores for URL segments, then we get a mix of underscores and hyphens which is less than ideal
- google uses hyphens to separate words, so if we ever want things to be searchable, hyphens tend to be better in terms of understandability for the search engine

PR #66, opened 2018-06-19
Accepting and rejecting a team invite

kntsoriano:
Closes https://github.com/exercism/reboot/issues/237.

## Description
This PR implements accepting and rejecting a team invite.

When a team invite is accepted, a user is added to the team and the team invite is destroyed.
When a team invite is rejected, the team invite is destroyed.

## Screenshot
screen shot 2018-06-19 at 1 32 02 pm


db/schema.rb
```diff
@@ -179,17 +179,17 @@
     t.index ["user_id"], name: "fk_rails_9f02fc96a0"
   end
 
-  create_table "repo_update_fetches", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci" do |t|
+  create_table "repo_update_fetches", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
iHiD: You've leaked some stuff here that you probably don't want to leak :)

config/routes.rb
```diff
@@ -134,6 +134,11 @@
     end
     resources :team_memberships, only: [:destroy]
 
+    resources :invitations, only: [] do
```
iHiD: As per #67, I'm not sure of why we have this as a separate routeset to the normal `invitations` one. What's your logic?

db/schema.rb
```diff
@@ -226,6 +226,7 @@
     t.integer "num_reactions", default: 0, null: false
     t.text "reflection"
     t.boolean "is_legacy", default: false, null: false
+    t.boolean "boolean", default: false, null: false
```
iHiD: You didn't want this either :)

app/views/teams/memberships/index.html.haml
```diff
@@ -6,7 +6,7 @@
       .section
         %h2 Pending Invitations
         %p These are people who have been invited to this team.
-        %table
+        %table.teams-table
```
iHiD: Are you extracting this because you're going to add the component for memberships below?

app/assets/stylesheets/teams/teams-table.scss
```diff
@@ -0,0 +1,27 @@
+@import "variables";
```
iHiD: Let's discuss this extraction on Slack. You seem to like to put things in seperate files. I like nesting stuff. Let's understand why we both think what we think :)

config/routes.rb
kntsoriano: I've left my explanation in the other PR.

app/views/teams/memberships/index.html.haml
kntsoriano: Yes. We have tables in two pages. I've extracted them because I'd want them to look the same.

app/assets/stylesheets/teams/teams-table.scss
kntsoriano: Personally, I'm a fan of putting things in separate files in order to slowly build our design components. Personally, it's been hard for me to add CSS because we don't have any styleguide which contains our defined components. I assume this is going to be hard for other people to contribute to as well, as for people new the codebase, we mostly just copy paste components that we see on other pages in order to build a new page.

app/views/teams/memberships/index.html.haml
iHiD: Ah - I'd missed they were in two different pages. Sorry.

app/assets/stylesheets/teams/teams-table.scss
iHiD: >  I'm a fan of putting things in separate files in order to slowly build our design components. 

I agree with this.

> Currently, it's been hard for me to add CSS because we don't have any styleguide which contains our defined components.

This is because we don't have any defined components :)

I think the problem really is that until you get to an MVP, you don't really know what the common components are. Many things look the same, but through the development process diverge enough that they're not really the same. I find that coding CSS as pages and then refactoring out components at the end is a much more effective and efficient way to do this than to try and refactor earlier. That's probably why you feel the CSS codebase is hard to get into at this stage. I find the process very different when refactoring code, where things tend to become cleaner not messier through the development process

There is a definite phase of refactoring / cleaning the CSS up, but it's not something I consider worth doing until we understand what we are keeping/removing/changing.

Does that make sense?

iHiD: For example, I'd consider the `inline-list.scss` breakout to be a premature refactor. There are dozens of different inline lists in the product that follow different rules at the moment. Once everything is at a finished stage, we can then start to go through and refine and polish to extract those components out and standardise them, but doing that at the moment would mean there were always risks to side-effects.

Thinking outloud, the side-effect thing is the big problem with all this, because it's not possible to test the way things look using code, and it's not reliable or realistic to go through every page checking every style of every state before deploy. **So reducing side-effects is the tradeoff I make for duplication.** That's the key thing for me in this, and maybe where you disagree?

kntsoriano: Looking back at it, I agree that `inline-list.scss` is a premature refactor because we only have one of it right now. You're correct that we should start refactoring once we get at least 3 three instances of the same component.

Looking back to why I did this, I found it hard to trace the CSS when I was only looking at element names rather than class names. I also needed to figure out the nesting in order to determine which style would apply. I think we can get to a middle point where we put our components in page-specific CSS files, but we start adding class names to them, in order to make it easier for us to trace the styles.

PR #67, opened 2018-06-19
Remove team invite

kntsoriano:
Closes https://github.com/exercism/reboot/issues/239.

## Description
This PR allows admins to remove team invites. Only admins are allowed to do this action.


config/routes.rb
```diff
@@ -129,6 +129,8 @@
     get "/" => "pages#index"
     get "dashboard" => "dashboard#index"
 
+    resources :invitations, only: [:destroy]
```
iHiD: Below this we have ` resources :invitations, only: [:new, :create, :destroy]`. I'd rather not duplicate that and just keep the original one. This also means you don't need to do the `skip_before_action`, because the team_id will be in the url.

app/controllers/teams/invitations_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Teams::InvitationsController < TeamsController
+  skip_before_action :find_team, only: [:destroy]
```
iHiD: See comment in routes.

I would rather add an extra scope which is `before_action :check_admin, only: [:create, :destroy]` and then do something like `current_user.managed_teams.where(id: params[:team_id]).exists?` there. It feels more clear of intention from a security POV than nesting that check within the methods.

app/views/teams/memberships/index.html.haml
```diff
@@ -9,9 +9,16 @@
...
-              %td= invitation.email
+              %td
+                = invitation.email
               -if current_team_membership.admin?
-                %td= link_to "Resend invite", "#"
+                %td
+                  %ul.inline-list
+                    %li= link_to "Resend invite", "#"
+                    %li= link_to "Remove",
+                      teams_invitation_path(invitation.id),
```
iHiD: See comment in routes.rb. And then this can change to `teams_team_invitation_path(@team.id, invitation.id)`

config/routes.rb
kntsoriano: I'm generally a fan of shallow routing in order to clearly see the intention behind the URL. If we're going to destroy an invitation, we wouldn't need to the team ID.

kntsoriano: Also, the `skip_before_action` was there because I needed to put it in. Personally, I feel that the `TeamsController` has too many responsibilities. It retrieves the teams, at the same time provides the layout for the teams page, but there are routes for the teams pages that don't really require a team, such as joining one. There should be a separate controller for each responsibility.

app/controllers/teams/invitations_controller.rb
kntsoriano: Agree with this.

config/routes.rb
iHiD: > I'm generally a fan of shallow routing in order to clearly see the intention behind the URL. If we're going to destroy an invitation, we wouldn't need the team ID.

OK - that makes sense. 

> Personally, I feel that the TeamsController has too many responsibilities. It retrieves the teams, at the same time provides the layout for the teams page, but there are routes for the teams pages that don't really require a team, such as joining one. 

I'm presuming you're talking about the `::TeamsController`, not the `Teams::TeamsController`? Yeah - I see your point here too. How would you structure it differently? Move `find_team` into `Teams::TeamsController` and have all the other controllers (e.g. `Teams::SolutionsController` inherit from `Teams::TeamsController` instead? That would make sense as a change to me.

PR #69, opened 2018-06-20
Updated honeycomb render

jackhughesweb:
I have updated the rendering of the track honeycomb display to avoid showing rows with small numbers of items, and added support for centering rows which are not completely filled.

This also updates some of the logic used in the calculation of the gopher's pupils (to restrict them to locations in the eye only)


iHiD:
Got a screen-capture of the gopher I could check so I don't have to download the branch? :)

jackhughesweb:
screen shot 2018-06-22 at 22 18 04
screen shot 2018-06-22 at 22 18 10

iHiD:
Great. Thanks @jackhughesweb :) Deploying it to the mentors-beta site now

PR #71, opened 2018-06-21
Add track information pages

iHiD:
I've written this then refactored it via meta-programming this. I think it's all fine as a first pass, but let me know if there's anything that raises alarm-bells pls. Thanks :)

Pages look like this:

screen shot 2018-06-21 at 15 35 08
....
screen shot 2018-06-21 at 15 35 12

---

The links on the new solution pages look like this:
screen shot 2018-06-21 at 15 36 38

### DIFF
--- a/app/controllers/track_pages_controller.rb
+++ b/app/controllers/track_pages_controller.rb
@@ -1,0 +1,16 @@
+class TrackPagesController < ApplicationController
+  before_action :get_track
+
+  Git::ExercismRepo::PAGES.each do |page|
+    define_method page do
+      @page = page
+      @content = ParsesMarkdown.parse(@track.repo.send(page))
+      render action: "show"
+    end
+  end
+
+  private
+  def get_track
+    @track = Track.find(params[:track_id])
+  end
+end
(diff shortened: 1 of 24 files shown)


kytrinyx:
Do we need "about the C track"? That should be included on the main track page already. (Or does the "about" link to the main track page?)

app/controllers/track_pages_controller.rb
```diff
@@ -0,0 +1,16 @@
+class TrackPagesController < ApplicationController
+  before_action :get_track
```
kytrinyx: Nit: `get_` is not a common prefix in Ruby. I would typically call this `track`

app/models/git/exercism_repo.rb
```diff
@@ -1,4 +1,5 @@
 class Git::ExercismRepo < Git::RepoBase
+  PAGES = %w{about installation tests learning resources}
```
kytrinyx: I mentioned this in the main conversation, but I would remove `about` here, since it's already included as part of the main track page.

kntsoriano:
Can we write a system test for this?

iHiD:
@knstoriano Yes, valid point. I didn't change any of the stuff around getting about.md from git so there might already be system tests for that. I'm also not sure how we check out the other repos on CI to test it, etc. Maybe create a reboot issue and add it as a to-do once we've got all the blocking stuff done?

PR #75, opened 2018-06-26
Add common-lisp file mapping to prism.

tushartyagi:
This PR adds the missing common-lisp mapping to the highlighter. Raised for https://github.com/exercism/v2-feedback/issues/170


iHiD:
Thanks. Deploying now :)

PR #76, opened 2018-06-26
Change core exercise unlocking

kntsoriano:
Finishes part of https://github.com/exercism/reboot/issues/123.

## Description
This PR changes how we unlock the next core exercise.

1. Don't unlock the next core exercise when there are still other core exercises in progress.
2. Unlock the core exercise with the lowest position that hasn't been completed yet.

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -29,6 +29,10 @@ class Exercise < ApplicationRecord
     !core
   end
 
+  def unlocked_by_user?(user)
+    solutions.find_by(user: user).present?
+  end
+
   def topic_names
     @topic_names ||= topics.pluck(:name).map(&:downcase)
   end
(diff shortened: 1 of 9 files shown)


app/models/exercise.rb
```diff
@@ -20,6 +20,17 @@ class Exercise < ApplicationRecord
   scope :core, -> { where(core: true) }
   scope :side, -> { where(core: false) }
 
+  scope :locked_for, -> (user) {
```
iHiD: This might be cleaner (and faster to execute) if refactored to:
```
where.not(id: user.solutions.select(:exercise_id)).
where(core: true)
```

app/models/exercise.rb
```diff
@@ -29,6 +40,10 @@ def side?
     !core
   end
 
+  def unlocked_by_user?(user)
+    solutions.find_by(user: user).present?
```
iHiD: This will be faster too as it won't retrieve the record from the db or have to create any AR objects:
`solutions.where(user_id: user.id).exists?`

PR #77, opened 2018-06-26
Update prism to use `lisp` instead of `common-lisp`

tushartyagi:
Prism supports `lisp` as a language and not `common-lisp`. This commit adds that change


iHiD:
Thanks!

PR #78, opened 2018-06-27
Show exercises in order

kntsoriano:
Closes https://github.com/exercism/reboot/issues/123.

## Description
This shows the exercises in the track page in order. The order of the exercises is the ff:

1. Completed
2. In progress
3. Locked

## Screenshots
screen shot 2018-06-27 at 12 20 13 pm


app/controllers/my/tracks_controller.rb
```diff
@@ -43,7 +43,23 @@ def show
     else
       core_exercises, side_exercises = normal_exercises.partition {|e|e.core?}
 
-      @core_exercises_and_solutions = core_exercises.map{|e|[e, mapped_solutions[e.id]]}
+      core_exercises = core_exercises.map do |e|
+        ExerciseWithSolution.new(e, mapped_solutions[e.id])
+      end
+
+      @core_exercises_and_solutions = core_exercises.
+        inject([[], [], []]) do |collection, exercise|
```
iHiD: I wonder if this would be nicer with `each_with_object` instead of inject so that we don't have to do L60.

app/controllers/my/tracks_controller.rb
```diff
@@ -43,7 +43,23 @@ def show
...
+          if exercise.completed?
+            collection[0] << exercise
+          elsif exercise.unlocked?
+            collection[1] << exercise
+          elsif exercise.locked?
+            collection[2] << exercise
+          end
+
+          collection
+        end.
```
iHiD: And maybe do this as `{}` rather than `do end` to make the `end.` a little nicer?

app/models/exercise_with_solution.rb
```diff
@@ -0,0 +1,25 @@
+class ExerciseWithSolution < SimpleDelegator
```
iHiD: This is fine for this PR, but I wonder if we should create a new directory to move this into and keep `models` as `AR Models`. What do you think?

kntsoriano: I don't have any preference. Some teams do with a separate `decorators/` and some don't. Let's see what works for us. I'll keep it in here for the meantime.

PR #79, opened 2018-07-01
Allow user to leave a track

iHiD:
Closes https://github.com/exercism/reboot/issues/72.

## Description
This PR allows a user to leave and rejoin a track. As we want to support rejoining a track, an `archived_at` attribute was added to `UserTrack` in order to determine whether a user has temporarily left the track. No data will be deleted and rejoining the track will simply be turning `archived_at` to `nil`.

### DIFF
--- a/app/controllers/my/user_tracks_controller.rb
+++ b/app/controllers/my/user_tracks_controller.rb
@@ -1,7 +1,15 @@
 class My::UserTracksController < MyController
   def create
     track = Track.find(params[:track_id])
-    JoinsTrack.join!(current_user, track)
+
+    if current_user.previously_joined_track?(track)
+      user_track = current_user.user_tracks.find_by(track: track)
+
+      user_track.update!(archived_at: nil)
+    else
+      JoinsTrack.join!(current_user, track)
+    end
+
     redirect_to [:my, track]
   end
 
--- a/app/controllers/my/user_tracks_controller.rb
+++ b/app/controllers/my/user_tracks_controller.rb
@@ -18,4 +18,11 @@ class My::UserTracksController < MyController
     redirect_to [:my, track]
   end
 
+  def leave
+    user_track = current_user.user_tracks.find(params[:id])
+
+    user_track.update(archived_at: Time.current)
+
+    redirect_to my_tracks_path
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -81,6 +81,13 @@ class User < ApplicationRecord
     false
   end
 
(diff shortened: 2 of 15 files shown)


kntsoriano:
@iHiD Please review. I couldn't add you as a reviewer.

app/controllers/my/user_tracks_controller.rb
```diff
@@ -18,4 +18,11 @@ def set_independent_mode
     redirect_to [:my, track]
   end
 
+  def leave
+    user_track = current_user.user_tracks.find(params[:id])
+
+    user_track.update(archived_at: Time.current)
```
iHiD: Add `!`

config/routes.rb
```diff
@@ -97,6 +97,7 @@
       member do
         patch :set_normal_mode
         patch :set_independent_mode
+        post :leave
```
iHiD: Two questions:
1) Do we want to use `destroy` rather than making a new action
2) If "no" to 1, do we want to set this to `delete :leave` instead of `post :leave`

app/models/user.rb
```diff
@@ -81,6 +81,13 @@ def may_view_solution?(solution)
...
 
+  def previously_joined_track?(track)
+    user_tracks.
+      archived.
+      where(track_id: track.id).
+      exists?
+  end
+
   def joined_track?(track)
     user_tracks.where(track_id: track.id).exists?
```
iHiD: I think this should have `unarchived` here.

app/views/my/tracks/_change_to_independent_mode_modal.html.haml
```diff
@@ -0,0 +1,10 @@
+=hex_green_track_icon(track)
+.main-section
+  %h2 Switch to Independent Mode
+  %h3 on the #{track.title} Track
+
+  %p In Independent Mode all of the exercises within a Track are unlocked and students can work through them freely. Note that in this mode Mentor feedback is disabled. This mode is for people who want to do fast deep-dives into a track and is not recommended for those that want the full Exercism experience.
+  .buttons
+    / TODO - kntsoriano
```
iHiD: Can you remove this TODO pls? I put it in the wrong file 😨

iHiD:
@kntsoriano I've made 4 comments :)

config/routes.rb
kntsoriano: I'm really unsure. I just feel that using a `delete` action has extra meaning that it deletes a record, and using a `post` is more neutral and can mean any action.

PR #80, opened 2018-07-04
Add C++ files mapping to prism.

pacmancoder:
This change will resolve https://github.com/exercism/v2-feedback/issues/194
Besides mapping of `.cpp` and `.hpp` I propose to map `.h` files as cpp header as it will produce normal syntax highlighting for c headers in most cases - I think it is better to not do any highlighting on .h at all?


iHiD:
@pacmancoder Thank you for this - very helpful!
@kytrinyx Do you have an opinion on the `.h` files. I imagine this is a common extension across languages?

kytrinyx:
I'm sorry I missed this. I think that if we're not taking the track into account when selecting a syntax highlighting scheme it might be best to stick with no highlighting on the `.h` files. I don't feel very strongly about it, though. At the moment I think the only tracks we have that would have `.h` files are C, C++, and Objective-C, and my suspicion is that they would be fairly similar to each other and that picking `c` to highlight them all might be fine.

iHiD:
@kytrinyx OK - shall I/you merge this now and we can re-address it later if we want to?

pacmancoder:
Ping @kytrinyx :ping_pong:

PR #81, opened 2018-07-05
Substitute configure command token in the walkthrough

kntsoriano:
Closes https://github.com/exercism/reboot/issues/257.

## Description
This PR changes the `[CONFIGURE_COMMAND]` placeholder in the walkthrough to a personalized token install command.


Gemfile
```diff
@@ -36,7 +36,7 @@ gem 'bootsnap', '>= 1.1.0', require: false
 
 group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
-  gem 'capybara', '~> 2.13.0'
+  gem 'capybara'
```
iHiD: Is it worth specifying that there's a requirement for `~3.3` here?

test/test_helper.rb
```diff
@@ -16,6 +16,17 @@ class ActionView::TestCase
...
+
+  def stub_repo_cache!(&block)
+    git_repo_cache = "#{Rails.root}/test/tmp/git_repo_cache"
+    Git::RepoBase.
+      stubs(:repo_base_dir).
+      returns(git_repo_cache)
+
+    yield
+
+    FileUtils.rm_rf(git_repo_cache)
```
iHiD: Is it worth wrapping this in an `ensure`?

test/test_helper.rb
```diff
@@ -39,4 +50,15 @@ def repo_mock
...
+
+  def using_repo_cache!(&block)
+    git_repo_cache = "#{Rails.root}/test/tmp/git_repo_cache"
+    Git::RepoBase.
+      stubs(:repo_base_dir).
+      returns(git_repo_cache)
+
+    yield
+
+    FileUtils.rm_rf(git_repo_cache)
```
iHiD: Same here about the `ensure`

Gemfile
kntsoriano: Good catch. Let's use `~> 3.0` as specified [here](https://github.com/rails/rails/pull/30959#discussion_r181931451).

test/test_helper.rb
kntsoriano: Yes! Thank you :)

PR #82, opened 2018-07-06
Exercism Icon Emoji added to Slack

rpottsoh:
Let me know if you would like the name of the file changed and or would like the file placed in a different location.


iHiD:
@rpottsoh Thank you for this! I'm now using this emoji in lots of Slack rooms. 

I've not replied sooner because I've been trying to work out where this should go. I don't think this should go in the website (ie this repo) as it's not something an asset serve. I think the right place is probably https://github.com/exercism/website-icons 

What do you think about adding a folder there called "emoji" and having this in it as "exercism.png" (ie `website-copy:/emoji/exercism.png`)?

rpottsoh:
Your welcome.   All sounds good.  I'll see about closing this PR and opening another in website-icons with the filename you've suggested. 

Sent from my android device.

PR #83, opened 2018-07-07
Add onboarding

iHiD:
Closes https://github.com/exercism/reboot/issues/95

@kntsoriano Can you check the logic of my `before_actions` please. People shouldn't be able to get beyond the signup without onboarding, basically.

### DIFF
--- a/app/assets/stylesheets/pages/onboarding.css.scss
+++ b/app/assets/stylesheets/pages/onboarding.css.scss
@@ -4,37 +4,41 @@
 #onboarding-page {
     .lhs {
         padding-right:30px;
-        padding-top:20px;
+        padding-top:30px;
     }
     .rhs {
         padding-left:50px;
-    }
-    h1 {
-        @include font-size(25, 25);
-        font-weight:$bold;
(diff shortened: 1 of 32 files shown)


app/assets/stylesheets/pages/onboarding.css.scss
```diff
@@ -0,0 +1,69 @@
...
+@import 'mixins';
+
+#onboarding-page {
+    .lhs {
+        padding-right:30px;
+        padding-top:20px;
+    }
+    .rhs {
+        padding-left:50px;
+        img {
```
kntsoriano: Can we remove this empty block?

app/controllers/onboardings_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      return
+    end
+  end
+
+  def migrate_to_v2
+  end
+
+  def update
+    current_user.update(
+      accepted_privacy_policy_at: DateTime.now,
```
kntsoriano: Can we use `Time.current` to use Rails' timezone settings vs our server's time?

PR #93, opened 2018-07-10
Site context management

kntsoriano:
Closes https://github.com/exercism/reboot/issues/230.

## Description
This PR implements site context management correctly by the ff:

1. Preserve cookies across subdomains by specifying the length of the tld.
2. Storing site context inside cookies instead of sessions.
3. Redirecting to the correct signed in path based on the site context.

While I was doing this feature, a good parallel I drew upon was when changing the country inside an e commerce website. This works the same way.


ccare:
👍 from my perspective. If you think it's safe, go ahead and merge, and we can firefight on the servers.

PR #95, opened 2018-07-10
Close account

kntsoriano:
Closes https://github.com/exercism/reboot/issues/91.

## Description
This PR implements account deletion. This only provides a script to be ran in the console, it isn't on the frontend. To delete a user, run `user.destroy`.

The user record is deleted, as well as associated objects except the ff:

1. Discussion posts where they are a mentor.
2. Maintainer records where their user record is associated.

Since maintainers has already this built in mind, I only needed to change the discussion posts model with the ff steps:

1. Removing the `null` check for `user_id` in the databse.
2. Add `optional: true` to the `belongs_to :user` relationship.
3. Adding null values when rendering a discussion post with a nil user in the front end.

## Things to check for
I may have missed more pages where discussion posts were rendered.


iHiD:
This looks good. Thanks dude :) Could you add a note on the README that describes what deleting a user does and does not do pls. Just so we remember in 6months when we use this for the first time. You can probably copy+paste from the PR text.

PR #100, opened 2018-07-10
Delete the pharo track until we get the syncing fixed

kytrinyx:
The Pharo track does not yet have any exercises. The SyncsTrack code
assumes that the code has an array of exercises, and blows up when there
are none.

It is likely a very quick fix, but we have so many things on our minds
at the moment with launch, that it's simpler to just remove this one
problem and then we can put it back when we've sorted out the syncing.

FYI @iHiD @kntsoriano I'm merging this to unblock contributors.


kytrinyx:
Nevermind. @iHiD already has a branch for this!

PR #139, opened 2018-07-12
Fix N+1 queries

kntsoriano:


Gemfile
```diff
@@ -45,6 +45,7 @@ group :development, :test do
   gem 'capybara', '~> 3.0'
   gem 'selenium-webdriver'
   gem 'factory_bot_rails'
+  gem 'bullet'
```
iHiD: Should we put this just in `development` (the block below)?

PR #160, opened 2018-07-13
Corrected typo.

cmondorf:
To bear something in mind is to keep it in mind. To bare something in mind would be to undress something in mind ;)

Source: https://www.grammarly.com/blog/bear-vs-bare/


kytrinyx:
I thought I had fixed this :) It turns out I fixed the same typo in a _different_ file. ✨ 
Thanks for the fix!

PR #170, opened 2018-07-14
Render walkthrough in a separate page

kntsoriano:
Fixes https://app.bugsnag.com/exercism/exercism-v2/errors/5b3a48d9fa7d9500181c24f9.

## Description
Upon looking into Bugsnag, we noticed that users tried to open the walkthrough in a new tab. This PR does fixes that by allow the walkthrough to be viewed in a separate page.

### DIFF
--- a/app/assets/javascripts/modal.coffee
+++ b/app/assets/javascripts/modal.coffee
@@ -16,6 +16,7 @@ $('#modal-wrapper').click (e) ->
 
 
 window.showModal = (className, html, options = {}) ->
+  options = JSON.parse(options)
   $("#modal-wrapper").removeClass()
   $("#modal-wrapper").addClass("no-autoclose") if options['no_autoclose']
   $("#modal-wrapper #modal").removeClass().addClass(className)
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/cli-walkthrough.scss
```diff
@@ -4,6 +4,8 @@
   @include normal-font;
   @include font-size;
 
+  padding: 20px;
```
iHiD: Does this affect the padding in the modal?

app/assets/javascripts/modal.coffee
```diff
@@ -16,6 +16,7 @@ $('#modal-wrapper').click (e) ->
 
 
 window.showModal = (className, html, options = {}) ->
+  options = JSON.parse(options)
```
iHiD: Have you checked this doesn't break existing modals that used options?

iHiD: Ah - maybe there aren't any that don't use the helper. Never mind.

PR #172, opened 2018-07-14
Fix leave track modal

kntsoriano:
Closes https://github.com/exercism/exercism.io/issues/3826.

### DIFF
--- a/app/assets/javascripts/modal.coffee
+++ b/app/assets/javascripts/modal.coffee
@@ -15,8 +15,12 @@ $('#modal-wrapper').click (e) ->
                   $e.find("#modal").hasClass('autoclose')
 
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modal.coffee
```diff
@@ -15,8 +15,12 @@ $('#modal-wrapper').click (e) ->
                   $e.find("#modal").hasClass('autoclose')
 
 
-window.showModal = (className, html, options = {}) ->
-  options = JSON.parse(options)
+window.showModal = (className, html, options = null) ->
+  if options
```
iHiD: Can you do this in JS? `options = options ? JSON.parse(options) : {}`

PR #187, opened 2018-07-17
Add syntax highlighting for elisp

yilkalargaw:
add a syntax highlighting for elisp

### DIFF
--- a/config/initializers/prism.rb
+++ b/config/initializers/prism.rb
@@ -209,6 +209,7 @@ Exercism::PrismFileMappings = {
   "diff": %w{},
   "docker": %w{},
   "eiffel": %w{},
+  "elisp": %w{lisp},
   "elixir": %w{ex exs},
   "erlang": %w{erl hrl },
   "fsharp": %w{fs fsx},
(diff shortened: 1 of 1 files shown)


iHiD:
@yilkalargaw Thanks for this.

So looking at the file, currently [any files ending in `.lisp`]([https://github.com/yilkalargaw/website/blob/247947d60d397638e4fa1f9e0742144a449d88b9/config/initializers/prism.rb#L241]) are being rendered by the `lisp` Prism decorators. As such your change probably won't do anything (as lisp will override elisp alphabetically). 

- Do lisp and elisp share file extensions? 
- How different is the syntax between the two?

yilkalargaw:
lisp is a language with operator first syntax. It has many dialects but many people mean common lisp so the file extesion of common lisp is used in prism js. Emacslisp is based on common lisp and uses many functions such as setq, map, if, cond, defun, let,  etc.. so it is a good substitute for emacs.

iHiD:
@yilkalargaw Thanks for the explainations

> .The file extensions used by emacs lisp is .el 

So the line of code in the PR should be `"elisp": %w{el}` to achieve this :)

iHiD:
I've sorted that out for you. Thanks for your help :)

PR #204, opened 2018-07-22
Improve approval flow

iHiD:
This PR does a few things
- It adds a new notification for when someone's solution is approved
- It adds a corrosponding mailer
- It adds extra behaviour for when people approve without comments, such as properly assigning the solution to them.
- It changes mentor's comment notifications so they're cleared when a mentor looks at the iteration with the comment on, not a different iteration.
- It does some minor refactoring / tidy up.

Fixes https://github.com/exercism/exercism.io/issues/3880
Fixes https://github.com/exercism/exercism.io/issues/3981
Fixes https://github.com/exercism/exercism.io/issues/3978

Probably also fixes loads of other issues I've not managed to find at a quick grok.

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -27,9 +27,9 @@ module NotificationsHelper
       notification.about.user.avatar_url
 
     when 'solution_approved'
-      # Your own face
-      # about = solution
-      notification.about.user.avatar_url
+      # Mentor's face
+      # trigger = mentor
+      notification.trigger.avatar_url
     end
   end
 end
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -2,10 +2,33 @@ module NotificationsHelper
   def notification_image(notification)
     case notification.type
     when 'new_discussion_post'
+      # Face of the discussion post user
+      # trigger = discussion post
       notification.trigger.user.avatar_url
+
     when 'new_discussion_post_for_mentor'
+      # Face of the discussion post user
+      # trigger = discussion post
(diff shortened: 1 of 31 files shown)


app/helpers/notifications_helper.rb
```diff
@@ -2,11 +2,34 @@ module NotificationsHelper
   def notification_image(notification)
     case notification.type
     when 'new_discussion_post'
+      # Face of the discussion post user
```
kytrinyx: What does `Face of` mean?

kytrinyx: Oh. Sorry. It's obvious now.

iHiD: This method was so hard to make sense of without the comments. I'd really like to refactor the way we do this at some point, but it's tricky.

app/services/creates_mentor_discussion_post.rb
```diff
@@ -3,21 +3,22 @@ def self.create!(*args)
     new(*args).create!
   end
 
-  attr_reader :iteration, :user, :content, :discussion_post
-  def initialize(iteration, user, content)
-    super
+  attr_reader :mentor
+  def initialize(iteration, mentor, content)
+    @mentor = mentor
+    super(iteration, mentor, content)
```
kytrinyx: I really like the distinction between `author` and `mentor` rather than using `user` everywhere.

app/services/approve_solution.rb
```diff
@@ -0,0 +1,45 @@
+class ApproveSolution
```
kytrinyx: This is definitely the right time for this extraction 👍

PR #238, opened 2018-07-26
Use headless Chrome for system tests.

dougal:
Discussion with @iHiD about system tests led to wondering how to use Chrome headless, and thus this PR.


iHiD:
Boom. Thanks dude.

PR #277, opened 2018-07-31
[a11y] Fix menu navigation

unused:
Fix header keyboard navigation order and ensure logical (header) document structure.

Make menu accessible by keyboard navigation.


iHiD:
This was conflicting so I fixed it and put it at https://github.com/exercism/website/pull/317, which I'll merge as soon as it goes green. This is another great step forward. Thank you 💙

PR #299, opened 2018-08-03
Temporarily removing exercise time labels

nickborromeo:
Resolves https://github.com/exercism/exercism.io/issues/3804

Instead of deleting the code to show the length of exercises, I am just commenting it out for now so that once the data for `exercise.length` is set, this feature can be toggled on again.

I tried fixing the `exercise_length_word` method however looking at the data my local instance has a length of 1.

```
execises = Exercise.all.pluck(:length)
```


iHiD:
Bonza! Thanks :)

You have a Slack invite to our repo btw if you want to join :)

PR #323, opened 2018-08-08
4001 preselect existing tracks

strickinato:
According specifications by iHid in https://github.com/exercism/exercism/issues/4001

This PR:

1. Set any existing mentorships as checked when the page loads.
2. Add any newly ticked mentorships on submission
3. Delete any unchecked mentorships on submission

It also adds an integration test ensuring that a mentor adding or removing tracks can see what they're doing.

🙌 Go Mentors! 🙌

Closes exercism/exercism#4001


test/integration/mentor_configure_test.rb
```diff
@@ -0,0 +1,21 @@
...
+class MentorConfigureTest < ActionDispatch::IntegrationTest
+  test "mentors configuring can see what tracks they already use" do
+    user = create :user
+    user.confirm
+
+    track_one = create :track
+    track_two = create :track
+    user.mentored_tracks = [track_one]
+
+    sign_in! user
```
iHiD: As an FYI, you can do `sign_in!` and then use `@current_user`, rather than having to set up and confirm the user yourself.

app/controllers/mentor/configure_controller.rb
```diff
@@ -2,6 +2,7 @@ class Mentor::ConfigureController < ApplicationController
   before_action :authenticate_user!
 
   def show
+    @mentored_track_ids = Set.new(current_user.mentored_tracks.pluck(:id))
```
iHiD: What's the reason behind using `Set.new` here rather than leaving it as an array?

app/views/mentor/configure/show.html.haml
```diff
@@ -14,7 +14,9 @@
           -Track.all.order(:title).pluck(:title, :id).each do |track_title, track_id|
             .field
               =label_tag "track_id_#{track_id}" do
-                =check_box_tag "track_id[#{track_id}]"
+                - is_mentored = @mentored_track_ids.member?(track_id)
+                - check_box_value = "track_id[#{track_id}]"
+                =check_box_tag check_box_value, check_box_value, is_mentored
```
iHiD: I think this could be `=check_box_tag check_box_value, 1, is_mentored`.

I don't think the double use of `check_box_value` is necessary?

test/integration/mentor_configure_test.rb
```diff
@@ -14,8 +14,26 @@ class MentorConfigureTest < ActionDispatch::IntegrationTest
...
     assert_select "form" do
-      assert_select "input[name=?][checked=checked]", "track_id[#{track_one.id}]"
-      assert_select "input[name=?]", "track_id[#{track_two.id}]"
+      assert_track_selected(track_one.id)
+      assert_track_not_selected(track_two.id)
     end
+
+    # User unclicks their track and cliks the other path
+    # effectively switching which tracks they belong to
+    put mentor_configure_path("track_id" => { track_two.id.to_s => true })
```
iHiD: What do you think about also adding a system test to test the form works?

Something like (this won't work but to give you clarity on what I mean):
```
get mentor_configure_path
click("input[name=track_id[#{track.id}]]")
click_button("Continue")
assert_equal [track], user.mentored_tracks 
```

iHiD:
@strickinato This is great, thanks! I've made a few minor questions/suggestions. I'd happily merge this as is so tell me if you want me to change/add anything I've mentioned or if you'd just like me to merge. I'm 100% cool either way :)

iHiD:
I've also added `Closes exercism/exercism#4001` to the PR description. This means that issue will automatically get closed when I merge this. You might know this and just decided you didn't think it was appropriate, but I thought I'd let you know in case you hadn't discovered it yet :)

strickinato:
> I think this could be =check_box_tag check_box_value, 1, is_mentored.

You we're totally right! I'm not in love with the "API" of this method

> What's the reason behind using Set.new here rather than leaving it as an array?

No good reason aside from my brain thinking about membership. But `Array#include?` is probably better given the small number of elements

> What do you think about also adding a system test to test the form works?

Aha!! This is actually what I was looking for!! I'm not used to what this code base calls "system" tests being called "integration tests" and when I was looking for testing, I didn't think to look in the system directory.

I've moved all the testing there because I think it does a much better job of covering what we want.


Thanks for your feedback. If you like the new tests. I'm very happy to get this merged!

strickinato:
And as a final note regarding: "I've also added Closes exercism/exercism#4001 to the PR description. "

I did know about that feature, and I'm all for computer automation. However this is a certain pleasure in manually closing an issue that I like to keep.

iHiD:
This is great. Merged. Thank you 💙

PR #325, opened 2018-08-08
Consider mentor count when rendering mentor info on the track page

ZapAnton:
Just a small change. 
If the current track has 1 mentor `Mentor` will be rendered instead of `Mentors`

### DIFF
--- a/app/views/tracks/_overview.html.haml
+++ b/app/views/tracks/_overview.html.haml
@@ -4,11 +4,7 @@
       .pure-u-1-3
         .icon
           =image "track-page-mentors.png", "Track mentors"
-        %h3 #{number_with_delimiter @track.mentors.size}
-        -if @track.mentors.size == 1
-          Mentor
-        -else
-          Mentors
+        %h3 #{number_with_delimiter @track.mentors.size} #{"Mentor".pluralize(@track.mentors.size)}
         .info Our mentors are friendly, experienced #{@track.title} developers who will help teach you new techniques and tricks.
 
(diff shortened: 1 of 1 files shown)


app/views/tracks/_overview.html.haml
```diff
@@ -4,7 +4,11 @@
       .pure-u-1-3
         .icon
           =image "track-page-mentors.png", "Track mentors"
-        %h3 #{number_with_delimiter @track.mentors.size} Mentors
+        %h3 #{number_with_delimiter @track.mentors.size}
+        -if @track.mentors.size == 1
```
iHiD: This needs to be indented below the h3. Currently this gives html of `<h3>5</h3> mentors` rather than `<h3>5 mentors</h3>`. 

You can also use the [pluralize](https://api.rubyonrails.org/classes/String.html#method-i-pluralize) helper to make life easier here. Could you maybe change this to something like:

```haml
%h3 #{number_with_delimiter @track.mentors.size} #{"Mentor".pluralize(@track.mentors.size)}
```

ZapAnton:
Done, @iHiD 
I also wanted to ask, if there is a specific format for commits on the `website` repo to specify a place where I make changes, like on the tracks repos, I would write `slug: Changes made to "slug"`

iHiD:
@ZapAnton No, not really. As long as the commit message makes sense I'm happy. I tend to squash and reword most PRs anyway. So here I'll squash them but would use your initial commit message.

PR #331, opened 2018-08-09
Remove default scope on exercise model and make it part of model relation

PatrickMcSweeny:
According to some, [default_scope is evil](https://rails-bestpractices.com/posts/2013/06/15/default_scope-is-evil/). In any case, I think it would make sense to remove the default scope from the exercise model and instead add the scope to the has_many :exercises relation on the track model. [This test](https://github.com/exercism/website/blob/master/test/system/exercises_test.rb#L4) already ensures that the exercises are being ordered correctly so I don't think this change would necessitate any new tests.

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -15,7 +15,6 @@ class Exercise < ApplicationRecord
   has_many :solutions
   has_many :iterations, through: :solutions
 
-  default_scope -> { order('position ASC, title ASC') }
   scope :active, -> { where(active: true) }
   scope :core, -> { where(core: true) }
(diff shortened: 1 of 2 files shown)


iHiD:
@PatrickMcSweeny I like this commit but I want to look over the codebase to make sure it won't have any side-effects. I don't trust that I've used `track.exercises` everywhere :)

I'll have a look when I get chance. Thanks! :)

iHiD:
@PatrickMcSweeny I've been through all instances of `Exercise.` and none of them seem to rely on ordering, so I'm going to merge this. It should give a performance boost to all of those cases too. Thank you :)

PR #335, opened 2018-08-11
User empty? method instead of comparing to zero

PatrickMcSweeny:
Using the any? and empty? methods on activerecord result sets and collections is better for performance and readability compared to comparing the length to zero.

### DIFF
--- a/app/controllers/api/solutions_controller.rb
+++ b/app/controllers/api/solutions_controller.rb
@@ -77,7 +77,7 @@ class API::SolutionsController < APIController
           exercises = Exercise.side.where(unlocked_by: nil).where(track_id: current_user.tracks).where(slug: params[:exercise_id]).includes(:track)
           if exercises.size == 1
             solution = CreateSolution.(current_user, exercises.first)
-          elsif exercises.size == 0
+          elsif exercises.empty?
             return render_404(:exercise_not_found)
           elsif exercises.size > 1
             return render_error(400, :track_ambiguous, "Please specify a track id", possible_track_ids: solutions.flat_map {|s|s.exercise.track.slug}.uniq)
--- a/app/controllers/my/settings_controller.rb
+++ b/app/controllers/my/settings_controller.rb
@@ -20,7 +20,7 @@ class My::SettingsController < MyController
 
     errors.uniq!
 
-    if errors.size == 0
+    if errors.empty?
       flash.notice = "Your track settings have been updated"
     elsif errors.size == 1
       flash.alert = "The handle #{errors[0]} is already taken"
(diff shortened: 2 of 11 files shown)


iHiD:
@PatrickMcSweeny Thanks for this :)

`any?` and `empty?` aren't actually better in all scenarios. `any?` will automatically load records, whereas `size` will do a `count(:all)` query in the database if the records aren't loaded. You can see this in the method defintions for [any?](https://apidock.com/rails/ActiveRecord/Relation/any%3F) and [size](https://apidock.com/rails/ActiveRecord/Relation/size).

 I'll need to go through and check for changes to SQL in these different states before merging. 

In the meantime, could you rebase/clean up the branch so only the one commit is on it pls?

PatrickMcSweeny:
Down to one commit.

iHiD:
@PatrickMcSweeny So I've read around this and `any?` is significantly slower than checking `size == 0` for arrays. For example, [this blog post has some benchmarks](https://mensfeld.pl/2015/06/empty-vs-blank-vs-any-why-you-should-not-use-any-to-test-if-theres-anything-in-the-array/).

It also has a different meaning. From the [Ruby docs](http://ruby-doc.org/core-2.4.0/Enumerable.html#method-i-any-3F): 
>Ruby adds an implicit block of { |obj| obj } that will cause any? to return true if at least one of the collection members is not false or nil.`

So for example:
```ruby
2.4.4 :004 > ![nil].empty?
 => true 
2.4.4 :005 > [nil].any?
 => false ```
```

So. I'd be happy if the change was the additions of `empty?` and `present?` as long as the records are loaded (which I think is the case in all of the areas you've changed). Want to change it to that?

PatrickMcSweeny:
Would it be worthy using exits? instead of any?

iHiD:
@PatrickMcSweeny `.exists?` will fire a SQL query to the db to ask if things exist. That's great if we're trying to find out if a record exists, but if we're actually going to use the records, then we've just doubled the number of queries for no gain, which will be a net performance loss.

So I think `.present?` is best generally if we're going to use the records if there are some.

PatrickMcSweeny:
screenshot from 2018-08-12 11-39-07

I did some experimenting in the console and it appears that when working with a result set assigned to a variable, it appears that present? fires an extra query while any? does not.

iHiD:
@PatrickMcSweeny `rails c` doesn't really do the same as a rails application because it executes ActiveRelations at the end of each line unless. To replicate what would happen in an app, you need to keep things on one line and not execute the command at the end of the line (using `;0` is good for that).

For example, this replicates an actual application call, and you'll see the `any?` version makes two calls, whereas `.present?` makes one.

```
es = Exercise.limit(10);es.present?;es.each{};0
es = Exercise.limit(10);es.any?;es.each{};0
```

PatrickMcSweeny:
Done

iHiD:
Finally got there. Thanks @PatrickMcSweeny! :)

PR #341, opened 2018-08-14
Add better error messages to the API

iHiD:
I've implemented this using Rails `I18n` functionality. Each error code that the API returns has a corresponding message in the I18n file.

I think this implements my bit of https://github.com/exercism/exercism/issues/4164#issuecomment-412716159
Closes https://github.com/exercism/exercism/issues/4159

### DIFF
--- a/app/controllers/api/solutions_controller.rb
+++ b/app/controllers/api/solutions_controller.rb
@@ -9,7 +9,7 @@ class API::SolutionsController < APIController
     end
 
     unless current_user.may_view_solution?(solution)
-      return render_403(:user_may_not_download_solution, "You may not download this solution")
+      return render_solution_not_accessible
     end
 
     responder = API::SolutionResponder.new(solution, current_user)
@@ -27,7 +27,6 @@ class API::SolutionsController < APIController
 
       begin
         track = Track.find(params[:track_id])
-        user_track = UserTrack.where(user: current_user, track: track).first
       rescue
         return render_404(:track_not_found, fallback_url: tracks_url)
       end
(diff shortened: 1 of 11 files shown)


app/controllers/api/solutions_controller.rb
```diff
@@ -93,15 +93,13 @@ def latest
 
   def update
     begin
-      solution = SolutionBase.find_by_uuid!(params[:id], current_user.id)
+      solution = SolutionBase.find_by_uuid!(params[:id])
     rescue
-      # This covers both a non-existing solution and a solution
-      # belonging to someone else. We might want seperate messages
-      # but I'm choosing not to leak the existance of a solution
-      # at this stage, but unifying the errors.
```
iHiD: ^^ We're leaking this in other places anyway, so I've just removed this and am making the error messages consistent with the rest of the API

config/locales/en.yml
```diff
@@ -30,4 +30,13 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
+  api:
+    errors:
+      solution_not_found: "This solution could not be found"
+      solution_not_accessible: "You do not have permission to view this solution"
+      track_not_found: "The track you specified does not exist"
+      track_ambiguous: "Please specify a track id"
```
kytrinyx: Can we make this `ID` instead of `id`?

iHiD:
@kytrinyx I've updated this now. Note that there is not test-coverage for the `teams` path yet. I'll add that seperately.

config/locales/en.yml
```diff
@@ -30,4 +30,15 @@
...
 en:
-  hello: "Hello world"
+  api:
+    errors:
+      duplicate_iteration: "No files you submitted have changed since your last iteration"
+      exercise_not_found: "The exercise you specified could not be found"
+      resource_not_found: "This URL was not recognised"
+      solution_not_found: "This solution could not be found"
+      solution_not_accessible: "You do not have permission to view this solution"
+      solution_not_unlocked: "You have not unlocked this solution"
```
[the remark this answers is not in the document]
iHiD: Agree. Updated it. Will merge/deploy when green.

PR #342, opened 2018-08-15
Fix formatting errors in readme files

PatrickMcSweeny:
Since my previous pull request to get rid of duplicate headers was merged, a lot of the readme files have formatting errors on the website because of the presence of extra newline characters. For instance:
readme

I was able to fix this by calling split("\n") instead of the lines method to split the readme markdown into lines. This way the website doesn't end up with all of the formatting errors.

### DIFF
--- a/app/models/git/exercise.rb
+++ b/app/models/git/exercise.rb
@@ -8,7 +8,7 @@ class Git::Exercise
   end
 
   def instructions
-    lines = exercise_reader.readme.lines
+    lines = exercise_reader.readme.split("\n")
     lines.shift if /^#\s*#{@title}\s*$/.match? lines.first
     ParseMarkdown.(lines.join("\n"))
   end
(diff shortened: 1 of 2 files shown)


app/models/git/exercise.rb
```diff
@@ -8,7 +8,7 @@ def initialize(exercise, git_slug, git_sha)
   end
 
   def instructions
-    lines = exercise_reader.readme.lines
+    lines = exercise_reader.readme.split("\n")
     lines.shift if /^#\s*#{@title}\s*$/.match? lines.first
     ParseMarkdown.(lines.join("\n"))
```
Insti: I was wondering how this worked: 
`lines` doesn't remove the final newline from the strings, but `split` does.
Removing the "\n" from the join would also have fixed the problem.

test/models/git/exercise_test.rb
```diff
@@ -6,7 +6,7 @@ class GitExerciseTest < ActiveSupport::TestCase
 
     readme_lines = ["# barfoo", "### Something else", "Here"]
```
Insti: Note that the mock test data does not match the strings that would have been returned by `lines` but does match those returned by `split("\n")`

PR #345, opened 2018-08-17
Use GFM spec for parsing markdown

kntsoriano:
Closes https://github.com/exercism/website-copy/issues/342.

## Description
When parsing our markdown files, we should use what GitHub uses as close as possible. This PR implements that.

## Screenshots

List now starts with the specified number, instead of resetting to "1".
screen shot 2018-08-17 at 2 25 39 am

## Links
https://github.com/gjtorikian/commonmarker -> contains GitHub's markdown parser setup.


iHiD:
@kntsoriano Thanks for doing this <3

So one thing that confuses me is that `redcarpet` *is* [GitHub's markdown parser](https://blog.github.com/2011-04-19-rolling-out-the-redcarpet/). So I'm confused why we shift away from it. If that's the right thing though, I'm happy to trust your wisdom, but I'd like to understand it :) 

So a few questions:
1. How does this do syntax highlighting? Does it use Prism as a post-step?
2. Could you maybe add a test for syntax highlighting?
3. Have you tested this lots manually?
4. Does this fix any of:
  - https://github.com/exercism/exercism/issues/4240
  - https://github.com/exercism/exercism/issues/4224
  - https://github.com/exercism/exercism/issues/4188
  - https://github.com/exercism/exercism/issues/3890
  - https://github.com/exercism/exercism/issues/3970

app/services/parse_markdown.rb
```diff
@@ -1,6 +1,4 @@
-require 'redcarpet'
-require 'rouge'
-require 'rouge/plugins/redcarpet'
+require 'commonmarker'
```
iHiD: Is there a reason we're not requiring this directly in the Gemfile?

iHiD:
@kntsoriano I've tested this now with some examples and it's excellent. I'm merging this and going to try and work out what it's fixed. If you could take a look at the comments above too that'd be good )

PR #354, opened 2018-08-18
Add service to merge JS and ECMA

iHiD:
Things that need to happen before merging:

- [x] All JS solutions should point to the equivalent ECMA exercise
- [x] Check in the db that there are no JS exercises without equivalent ECMA slugs.
- [x] Update all shas to the magic sha that @kytrinyx provides me with.
- [x] Add user_track records 
- [x] Add track_mentorship records

I have checked there are no missing slugs by doing this:
```ruby
js_slugs = Track.find_by_slug('javascript').exercises.pluck(:slug).sort
ecma_slugs = Track.find_by_slug('ecmascript').exercises.pluck(:slug).sort
js_slugs - ecma_slugs

=> []

```

### DIFF
--- a/app/services/flux/merge_js_and_ecma.rb
+++ b/app/services/flux/merge_js_and_ecma.rb
@@ -8,7 +8,7 @@ module Flux
       lock_js_shas
       repoint_solutions
       rename_js_exercises
-      rename_tracks
+      update_tracks
       creates_track_mentorships
       fix_unlocking
     end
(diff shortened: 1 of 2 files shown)


kytrinyx:
Magic SHA: `6a8a5a41b89a45008b46ca18ff7ea800baca1c4c`

app/services/flux/merge_js_and_ecma.rb
```diff
@@ -0,0 +1,114 @@
...
+    end
+
+    private
+
+    def lock_js_shas
+      Solution.where(exercise_id: js.exercises).update_all(git_sha: JS_SHA)
+    end
+
+    def repoint_solutions
+      # Delete any unstarted ecma exercises
```
iHiD: These will get readded by the FixUnlocking script, but doing this will speed things up a lot.

PR #361, opened 2018-08-20
fix unclosed v1 migration modal

Cohen-Carlisle:
The v1_migration_modal was broken because the controller was returning
javascript to hide the .main-section and show the .start-section, but
the .start_section only exists on the started-modal.
By removing `remote: true` we will receive a redirect, effective closing
the modal.
It's not the most efficient solution, but it's cleaner than messing with
the js response to conditionally hide/show or close the modal.


Cohen-Carlisle:
Note: I didn't test this locally as this modal seems untested. However, I'm pretty confident this is fine as this is the way the [`change_to_mentored_mode_modal`](https://github.com/exercism/website/blob/master/app/views/my/tracks/_change_to_mentored_mode_modal.html.haml) works.

PR #364, opened 2018-08-20
Don't insert bullet point after horizontal rule in markdown editor

kntsoriano:
Closes https://github.com/exercism/exercism/issues/4188.

## Description
If we type in a horizontal rule in the markdown editor, a bullet point is inserted automatically after hitting Enter.

## Solution
The problem lies within our markdown editor library. I've submitted [a PR](https://github.com/toopay/bootstrap-markdown/pull/323) to fix this, but since we keep a local copy of the library, I decided to change it here, too. We can update our local copy to match the library once the PR gets merged in.

### Before
aug-20-2018 18-02-03
### After
aug-20-2018 18-08-37

### DIFF
--- a/app/assets/javascripts/libs/bootstrap-markdown.js
+++ b/app/assets/javascripts/libs/bootstrap-markdown.js
@@ -874,6 +874,10 @@
             }
           }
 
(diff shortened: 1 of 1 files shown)


iHiD:
Awesome. Great work @kntsoriano :D

PR #372, opened 2018-08-22
[WIP] Replicate Github setup for parsing markdown and syntax highlighting

kntsoriano:
**DO NOT MERGE FOR NOW**

## Description
This setup tries to get as close to what Github has for markdown parsing and syntax highlighting. I inferred their setup by looking at the [html-pipeline](https://github.com/jch/html-pipeline) gem.

I've noticed that we use Github syntax highlighting as our "source of truth" to determine whether we parse markdown and highlight tokens correctly. If we get closest to their setup, it would be really helpful. Also, using what they've done also reduces maintenance cost for us.

## Todos
1. As this doesn't use Prism anymore, we should look to remove it. I'm just unsure where the other parts of it are.


kntsoriano:
I'll close this for now, but I wont delete the branch :)

PR #382, opened 2018-08-30
Split settings page into two separate pages

jackhughesweb:
Changes include:
- Moving privacy options to a new page
- Replacing the navigation bar on each settings page with links to the other pages
- Removing the "Preferences" option in the drop down menu
- Adding tests for these pages

### DIFF
--- a/app/controllers/my/track_settings_controller.rb
+++ b/app/controllers/my/track_settings_controller.rb
@@ -1,0 +1,43 @@
+class My::TrackSettingsController < MyController
+  def show
+    setup_show
+  end
+
+  def update_user_tracks
+    @user_tracks = current_user.user_tracks.includes(:track)
+    mapped_user_tracks = @user_tracks.each_with_object({}) {|ut, h|h[ut.id] = ut }
+
+    errors = []
+    params[:user_tracks].each do |id, data|
+      anon = data[:anonymous].to_i == 1
+      handle = data[:handle].present?? data[:handle] : nil
+
+      success = mapped_user_tracks[id.to_i].update(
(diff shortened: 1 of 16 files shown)


app/views/my/settings/show.html.haml
```diff
@@ -1,15 +1,11 @@
...
-    .lo-container
-      /=link_to "Dashboard", root_path
-      /=nav_divider
-      %span Settings
+  =settings_bar
 
   = notice_and_alert(current_user)
   .lo-container
     %h1 Account settings
     .pure-g
```
iHiD: Can we get rid of this `pure-g` and `pure-1-1` altogether?

app/views/my/track_settings/show.html.haml
```diff
@@ -0,0 +1,56 @@
+#my-track-settings-page
+  =settings_bar
+
+  = notice_and_alert(current_user)
+  .lo-container
+    %h1 Track settings
+    .pure-g
```
iHiD: Here too.

app/controllers/my/track_settings_controller.rb
```diff
@@ -0,0 +1,43 @@
+class My::TrackSettingsController < MyController
+  def show
+    setup_show
+  end
+
+  def update_user_tracks
```
iHiD: Could we change this just to be `update`?

config/selenium.yml
```diff
@@ -0,0 +1,4 @@
+test:
```
iHiD: Could you rebase this out pls?

PR #387, opened 2018-09-04
Make command hint fields read only

PatrickMcSweeny:
Closes exercism/exercism#4329.


iHiD:
@PatrickMcSweeny Thanks :) Have you verified that the copy button still works?

PatrickMcSweeny:
@iHiD Yes I just checked again to make sure. Unfortunately there does not seem to be a way to do that through system tests though.

iHiD:
Great. Thanks. I've merged and am deploying :)

PR #389, opened 2018-09-04
Pluralize track page lo-container stats

ZapAnton:
For some reason the current track page rendered stats like Mentors or Maintainers are not pluralized (for example on the [Rust track page](https://exercism.io/my/tracks/rust) it is rendered `1 MAINTAINERS`)

This PR tries to address it with the changes similar to #325


iHiD:
@ZapAnton Thanks for this. The build failed if you'd like to take a look.

PR #421, opened 2018-10-06
Add test for signing out

PatrickMcSweeny:
Closes exercism/exercism#4216

### DIFF
--- a/test/system/signout_test.rb
+++ b/test/system/signout_test.rb
@@ -1,0 +1,13 @@
+require "application_system_test_case"
+
+class SignoutTest < ApplicationSystemTestCase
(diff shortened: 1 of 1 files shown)


iHiD:
Thanks. I'm very happy to merge this but I'm not sure it actually closes exercism/exercism#4216 as I still don't understand what caused that problem.

PatrickMcSweeny:
I think it was a matter of accidentally opening the 'sign out' link in a new tab, which results in a GET request being sent, when route is set up to accept a DELETE request.

iHiD:
Cool - that makes sense! Thanks :)

PR #441, opened 2018-10-22
Fix width of notification bar

PatrickMcSweeny:
Closes exercism/exercism#4420 and closes exercism/exercism#4435. Screenshot:

screenshot from 2018-10-22 18-35-40


iHiD:
Brilliant. Thanks :)

PR #454, opened 2018-11-07
Fix bug in switching from independent to mentoring mode

iHiD:
(cc @ErikSchierboom @sshine)

Note to self. Once this is deployed, run:

```ruby
User.find(231415).solutions.joins(:exercise).where("exercises.track": Track.find_by_slug("csharp")).submitted.not_completed.where("exercises.core": true).where.not(mentoring_requested_at: nil).where(approved_by_id: nil).order("position ASC").offset(1).update_all("mentoring_requested_at = NULL")
```

### DIFF
--- a/app/services/unlock_next_core_exercise.rb
+++ b/app/services/unlock_next_core_exercise.rb
@@ -4,7 +4,10 @@ class UnlockNextCoreExercise
   initialize_with :track, :user
 
   def call
-    return core_exercise_in_progress if core_exercise_in_progress
+    if existing_core_exercise
+      existing_core_exercise.update(mentoring_requested_at: Time.now) unless existing_core_exercise.mentoring_requested_at
+      return existing_core_exercise
+    end
 
     next_exercise = track.exercises.core.
                                     locked_for(user).
@@ -22,7 +25,7 @@ class UnlockNextCoreExercise
   private
 
   memoize
-  def core_exercise_in_progress
+  def existing_core_exercise
     user.
       solutions.
       joins(:exercise).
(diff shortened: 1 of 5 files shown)


app/services/unlock_next_core_exercise.rb
```diff
@@ -4,7 +4,10 @@ class UnlockNextCoreExercise
   initialize_with :track, :user
 
   def call
-    return core_exercise_in_progress if core_exercise_in_progress
+    if existing_core_exercise
+      existing_core_exercise.update(mentoring_requested_at: Time.now) unless existing_core_exercise.mentoring_requested_at
```
sshine: Does Rails cache `existing_core_exercise` between these two lines?

iHiD: Yeah, [here](https://github.com/exercism/website/pull/454/files/23e6145f067f1569063b1d93fee96d6e3cc2789b#diff-da68e877c63f14f0281db6cef96f9c9dR27).

It's part of the [Mandate gem](http://github.com/ihid/mandate)

Thanks for checking :)

sshine: Sorry I deleted my comment; I noticed `memoize` just below.

Or rather, I noticed `memoized` first, gave it no thought, asked the question, and *then* started thinking. ;-)

iHiD: Ha!

ErikSchierboom:
Quick fix! Thanks!

iHiD:
My pleasure. Thanks for highlighting it. Anything that means data is wrong gets my urgent attention, because otherwise future-iHiD will have data-fixing to do, and I hate that :)

PR #460, opened 2018-11-16
Remove N+1 queries from ExerciseMap

kntsoriano:
Fixes https://github.com/exercism/exercism/issues/4454.


app/models/exercise_with_solution.rb
```diff
@@ -3,9 +3,8 @@ class ExerciseWithSolution < SimpleDelegator
 
   delegate :approved?, :in_progress?, to: :solution
 
-  def initialize(exercise, solution)
+  def initialize(exercise)
```
iHiD: I think this should still take two arguments. If the class ever get reused the `exercise.solutions.first` thing would be unexpected behaviour.

app/models/exercise_map.rb
```diff
@@ -23,7 +23,10 @@ def map
   def exercises
     @exercises ||= track.
       exercises.
-      map { |e| ExerciseWithSolution.new(e, solution_for(e)) }
+      left_outer_joins(:solutions).
+      includes(:unlocks).
+      where("solutions.user_id = ? OR solutions.user_id IS NULL", user.id).
+      map { |e| ExerciseWithSolution.new(e) }
```
iHiD: In line with my comment below, I think this should be `ExerciseWithSolution.new(e, e.solutions.first)`

app/models/exercise_with_solution.rb
kntsoriano: Thanks! I agree that's a better solution.

PR #465, opened 2018-11-17
Allow non-dangerous tags in comments

iHiD:
(cc @kotp)


kotp:
We get our hints back!

PR #476, opened 2018-11-22
Show solution last_updated_by_user_at in number of days

PatrickMcSweeny:
Closes exercism/exercism#4374

### DIFF
--- a/app/helpers/mentor_dashboard_helper.rb
+++ b/app/helpers/mentor_dashboard_helper.rb
@@ -1,0 +1,12 @@
+module MentorDashboardHelper
+  SECONDS_PER_DAY = 60 * 60 * 24
+
+  def time_ago_in_words(time)
+    if (DateTime.now - time.to_datetime) * SECONDS_PER_DAY < 1.day
+      super(time)
+    else
+      number_of_days = ((Time.now - time) / SECONDS_PER_DAY).to_i
+      pluralize(number_of_days, 'day')
+    end
+  end
+end
--- a/test/helpers/mentor_dashboard_helper_test.rb
+++ b/test/helpers/mentor_dashboard_helper_test.rb
@@ -1,0 +1,27 @@
+require "test_helper"
+
+class MentorDashboardHelperTest < ActionView::TestCase
+  test "time_ago_in_words calculates time in days, not months" do
+    assert_equal "33 days", time_ago_in_words(Time.now - 33.days)
+  end
+
+  test "time_ago_in_words calculates time in days" do
+    assert_equal "2 days", time_ago_in_words(Time.now - 2.days)
+  end
+
+  test "time_ago_in_words calculates time less than one day in hours or minutes" do
+    assert_equal "about 2 hours", time_ago_in_words(Time.now - 2.hours)
+  end
+
+  test "time_ago_in_words correctly calculates one day plus one second" do
+    assert_equal "1 day", time_ago_in_words(Time.now - (SECONDS_PER_DAY + 1))
+  end
+
+  test "time_ago_in_words correctly calculates one day minus one second" do
+    assert_equal "1 day", time_ago_in_words(Time.now - (SECONDS_PER_DAY - 1))
+  end
+
+  test "time_ago_in_words correctly calculates one day minus one minute" do
+    assert_equal "about 24 hours", time_ago_in_words(Time.now - (SECONDS_PER_DAY - 60))
+  end
+end
--- a/test/system/mentor/dashboard_test.rb
+++ b/test/system/mentor/dashboard_test.rb
@@ -1,0 +1,28 @@
+require 'application_system_test_case'
+
+class DashboardTest < ApplicationSystemTestCase
+  test "displays the number of days ago a solution was updated in days, not months" do
+    some_date = Date.new(2019, 1, 1)
+    solution_updated_at = Time.now - 33.days
+
+    track = create(:track)
+    mentor = create(:user_mentor,
+                    accepted_terms_at: some_date,
+                    accepted_privacy_policy_at: some_date,
+                    mentored_tracks: [track])
+    exercise = create(:exercise, title: "Exercise 1", track: track)
+    solution = create(:solution, exercise: exercise,
+                mentoring_requested_at: solution_updated_at,
+                last_updated_by_user_at: solution_updated_at)
+    create(:iteration, solution: solution)
+    create(:solution_mentorship,
+           user: mentor,
+           solution: solution,
+           requires_action_since: solution_updated_at)
+
+    sign_in!(mentor)
+    visit mentor_dashboard_path
+
+    assert page.has_content?('33 days ago')
+  end
+end


PatrickMcSweeny:
I tried creating a cong/initializers/date_helper.rb file with the following contents:

`module ActionView
  module Helpers
    module DateHelper
      def time_ago_in_words(from_time, options = {})
        from_time = from_time.to_time
        if (Time.now.to_datetime - from_time.to_datetime) * 86400 < 1.day
          super(from_time)
        else
          "#{((Time.now - from_time) / 86400).to_i} days"
        end
      end
    end
  end
end
`
but I get the following error:

ActionView::Template::Error: super: no superclass method `time_ago_in_words'

PatrickMcSweeny:
I tried creating a config/initializers/date_helper.rb file with the following contents

`module ActionView`
`  module Helpers`
`    module DateHelper`
`      def time_ago_in_words(from_time, options = {})`
`        from_time = from_time.to_time`
`        if (Time.now.to_datetime - from_time.to_datetime) * 86400 < 1.day`
`          super(from_time)`
`        else`
`          "#{((Time.now - from_time) / 86400).to_i} days"`
`        end`
`      end`
`    end`
`end`

But I get the following error:

ActionView::Template::Error: super: no superclass method `time_ago_in_words'

Also, I'm not sure it would be a good idea to override the method globally, that might have some unintended consequences later on.

PatrickMcSweeny:
@iHiD any more thoughts on this?

app/helpers/mentor_dashboard_helper.rb
```diff
@@ -0,0 +1,10 @@
+module MentorDashboardHelper
+  def time_ago_in_days(time)
+    time = time.to_time
+    if (Time.now.to_datetime - time.to_datetime) * 86400 < 1.day
+      time_ago_in_words(time)
+    else
+      "#{((Time.now - time) / 86400).to_i} days"
```
kytrinyx: Are you familiar with the concept of "magic numbers" and why they're undesirable?

Here the number 86400 occurs twice, and there's no context as to what it means. The most common way to avoid magic numbers is to define them as well-named constants, and if possible defined with an expression that is obviously not wrong.

For example 86400 is the number of seconds in a day. You could say:

```
SECONDS_PER_DAY = 86400
```

But that still means that the reader has to double-check... is that really the number of seconds in a day?

Better would be:

```
SECONDS_PER_DAY = 60 * 60 * 24
```

The numbers 60 and 24 immediately speak to the reader, which means the reader can glance at it and move on without having to do a double-take.

test/helpers/mentor_dashboard_helper_test.rb
```diff
@@ -0,0 +1,13 @@
+require 'test_helper'
+
+class MentorDashboardHelperTest < ActionView::TestCase
+  test 'time_ago_in_days calculates time in days' do
+    two_days_ago = Date.today - 38.days
+    assert_equal '38 days', time_ago_in_days(two_days_ago)
```
kytrinyx: This is very confusing. The local variable is named `two_days_ago`, and there's a magic number `38`. Why 38? How is two days ago `38 days`?

As a rule of thumb, I use the simplest, stupidest values possible that will demonstrate that the behavior I want is correct. I also want to check boundary values.

So for this behavior I would want to check at least 1 and 2 (because of pluralization), and possibly 0 if that would say `0 days` since if we happened to not use the default Rails behavior here, we might end up with `0 day`.

test/helpers/mentor_dashboard_helper_test.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  test 'time_ago_in_days calculates time less than one day in hours or minutes' do
+    two_hours_ago = (Time.now - 2.hours).to_datetime
+    assert_equal time_ago_in_words(two_hours_ago), time_ago_in_days(two_hours_ago)
```
kytrinyx: This is very confusing. Why is it called `time_ago_in_days` if it will use hours or minutes too?
Where is the time_ago_in_words method from? Why is the assertion defined as being the same as the words method? How would the reader know that the words method is doing the right thing?

Generally test assertions should be against hard-coded, simple values, not against another implementation, unless the test is actually testing that two implementations are the same as part of a long-term refactoring, or something similar.

As a general rule, I would avoid using local variables if the name basically says the same thing as the implementation. `two_hours_ago` is no more readable than `Time.now - 2.hours`. That also avoids the problem that you have above when you end up with `two_days_ago = Date.today - 38.days` because you changed your mind about something. If you would have to change the name of the variable if you changed something trivial about the expression that you used to define it, that's typically a hint that the variable name is too specific or too concrete.

test/system/mentor/dashboard_test.rb
```diff
@@ -0,0 +1,25 @@
...
+                    mentored_tracks: [track])
+    exercise = create(:exercise, title: "Exercise 1", track: track)
+    solution = create(:solution, exercise: exercise,
+                mentoring_requested_at: Time.current,
+                last_updated_by_user_at: Time.now - 38.days)
+    create(:iteration, solution: solution)
+    create(:solution_mentorship,
+           user: mentor,
+           solution: solution,
+           requires_action_since: Date.yesterday)
```
kytrinyx: Here the key information that the reader is looking for is mixed into all the boilerplate. To make it as easy as possible for the reader of the test, I would ask myself what would be critical to know in order to read the assertion and be certain that it's testing the right thing?

I can't tell if the dates on lines 7 and 8 are important (probably not), and also I'm not sure why they're chosen as Christmas day. If they're not important I would pick something simpler, like January 1st. I might also put it in a local variable named:

```
some_date = Date.new(2016, 1, 1)
```

I also can't tell if it's important that the mentoring was requested right now (line 12), or when it requires action since (line 18). On the other hand, I'm puzzled at the fact that the solution was last updated 38 days ago, but only requires action since yesterday. I'm guessing that those two times are not important for the test, but I would try to avoid making the reader have to think too hard about it.

My understanding after having read through this, is that the only thing that _is_ important is the timestamp on line 13.
Since that timestamp is important, I would put it in a local variable at the very top.

That would make this something like:

```
solution_updated_at = Time.now - 38.days

unimportant_date = Date.new(2016, 1, 1)
unimportant_time = Time.current
track = create(:track)
mentor = create(:user_mentor, ...)
exercise = create(:exercise, ...)
solution = create(:solution, ...)
create(:iteration, ...)
create(:solution_mentorship, ...)

sign_in!(mentor)
visit mentor_dashboard_path

assert page.has_content?('38 days ago')
```

PatrickMcSweeny:
@kytrinyx Thank you for the detailed and thoughtful feedback. I have made updates based on your suggestions.

test/helpers/mentor_dashboard_helper_test.rb
```diff
@@ -0,0 +1,11 @@
+require "test_helper"
+
+class MentorDashboardHelperTest < ActionView::TestCase
+  test "time_ago_in_days calculates time in days" do
+    assert_equal "2 days", time_ago_in_days(Time.now - 2.days)
+  end
+
+  test "time_ago_in_days calculates time less than one day in hours or minutes" do
+    assert_equal "about 2 hours", time_ago_in_days(Time.now - 2.hours)
```
kytrinyx: I'm still not quite clear on why the method is called `time_ago_in_days` instead of `time_ago_in_words`. How would you feel about overriding Rails implementation of `time_ago_in_words`?

PatrickMcSweeny: I had tried earlier to override the method in an intializer file but this time I just did it in a regular helper file and it worked fine. I guess there is a risk in overriding a method globally but in this case I think the new implementation is an improvement so that would make it worthwhile.

kytrinyx: If we're overriding the method, we want to override it everywhere, not just in a subset of the project, as consistency is important.

test/helpers/mentor_dashboard_helper_test.rb
```diff
@@ -0,0 +1,11 @@
+require "test_helper"
+
+class MentorDashboardHelperTest < ActionView::TestCase
+  test "time_ago_in_words calculates time in days" do
+    assert_equal "2 days", time_ago_in_words(Time.now - 2.days)
+  end
+
+  test "time_ago_in_words calculates time less than one day in hours or minutes" do
+    assert_equal "about 2 hours", time_ago_in_words(Time.now - 2.hours)
+  end
```
kytrinyx: Let's get some tests around the boundary conditions.

Since the cutoff is `SECONDS_PER_DAY`, it would make sense to have a test for exactly `SECONDS_PER_DAY` ago, `SECONDS_PER_DAY - 1` seconds ago, and `SECONDS_PER_DAY + 1` seconds ago.

PatrickMcSweeny: It looks like Ruby rounds day minus one second up to a day, so I added another test to show that it will display a day minus one minute in hours (the actual cut off is 31 seconds).

kytrinyx: Neat, I didn't know that.

app/helpers/mentor_dashboard_helper.rb
```diff
@@ -0,0 +1,14 @@
+module MentorDashboardHelper
+  SECONDS_PER_DAY = 60 * 60 * 24
+
+  def time_ago_in_words(time)
+    case (DateTime.now - time.to_datetime) * SECONDS_PER_DAY
+    when 0..1.day
+      super(time)
+    when 1.day..2.days
+      "1 day"
+    else
```
kytrinyx: Since ActiveSupport is included in the project, this can use the `pluralize` inflector to simplify.
https://apidock.com/rails/String/pluralize

PatrickMcSweeny: Updated

kytrinyx:
@iHiD this is ready for final review from you.

app/helpers/mentor_dashboard_helper.rb
PatrickMcSweeny: I'm actually not sure the new approach is an improvement.

PatrickMcSweeny:
I added another to reflect the original point of overriding the rails implementation of time_ago_in_words, which was to display lengths in days rather than months.

app/helpers/mentor_dashboard_helper.rb
```diff
@@ -0,0 +1,12 @@
+module MentorDashboardHelper
+  SECONDS_PER_DAY = 60 * 60 * 24
+
+  def time_ago_in_words(time)
+    if (DateTime.now - time.to_datetime) * SECONDS_PER_DAY < 1.day
+      super(time)
+    else
+      number_of_days = ((Time.now - time) / SECONDS_PER_DAY).to_i
+      "#{number_of_days} #{'day'.pluralize(number_of_days)}"
```
iHiD: I believe this could be changed to

```suggestion
      pluralize(number_of_days, 'day')
```

PatrickMcSweeny: Done

app/helpers/mentor_dashboard_helper.rb
```diff
@@ -0,0 +1,12 @@
+module MentorDashboardHelper
+  SECONDS_PER_DAY = 60 * 60 * 24
+
+  def time_ago_in_words(time)
+    if (DateTime.now - time.to_datetime) * SECONDS_PER_DAY < 1.day
+      super(time)
+    else
+      number_of_days = ((Time.now - time) / SECONDS_PER_DAY).to_i
+      "#{pluralize(number_of_days, 'day')}"
```
kytrinyx: This doesn't need to be an interpolated string, the `pluralize()` method returns a string.

PatrickMcSweeny: Updated

iHiD:
Thank you both :)

PR #478, opened 2018-11-22
Remove requirement for team invitation email to match

iHiD:
We want to let someone accept a team invite even if the email is a mismatch, as someone might sign up with a different email. This allows someone to click on a link in an email, log in with a different account, and still accept it.


dougal:
Hype!

PR #483, opened 2018-11-24
Allow editing of public solutions

iHiD:
Closes https://github.com/exercism/exercism/issues/4163
Closes https://github.com/exercism/exercism/issues/4311


lloydchang:
@iHiD Will reflections be editable too? Thank you!

iHiD:
Yeah. We've got some UI for that designed. Needs a few hours of work (which I'm hard pushed to find atm) but I'll try and do it before the end of the year :)

PR #484, opened 2018-11-26
Fix  #4350 Code not lining up with line numbers

mrtnmgs:
I simply removed code that caused the issue. Unfortunately the code I removed was not commented so it is hard to understand why it was here in the first place. I didn't notice anything breaking when I removed it (I checked in latest versions of FF, Chrome, Safari and Opera on Mac OS 10.13.6). My guess is that it was fixing a bug with Prism which has been fixed since?

### DIFF
--- a/app/assets/stylesheets/prism-modifications.scss
+++ b/app/assets/stylesheets/prism-modifications.scss
@@ -8,18 +8,6 @@ pre.line-numbers {
     @include width-small() {
         font-size:100%;
     }
-    .line-numbers-rows {
-        span {
-            line-height:13px;
-            padding-bottom:1px;
-            @include width-small() {
-                line-height:16px;
-                padding-bottom:5px;
-            }
-            &:before {
-            }
-        }
-    }
     .line-highlight {
         background: rgba($color-2, 0.1);
         margin-top:0;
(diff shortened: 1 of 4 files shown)


PatrickMcSweeny:
The code was added by @jackhughesweb in [this commit](https://github.com/exercism/website/commit/fbe8806211b4034be67f34f553c09fe26e036940).

mrtnmgs:
Thanks @PatrickMcSweeny, I should have thought to check the file history!
@jackhughesweb do you happen to remember what your commit what fixing specifically? I removed the code and I'm not seeing any issues on mobile...

sshine:
This PR seems to have stalled due to missing response from @jackhughesweb. Is there an alternative path of action we can follow to assert that removing this CSS has no bad consequence? Perhaps we can just test it and see if other problems occur? Since right now this affects all solutions on the website of more than a few lines.

jackhughesweb:
@mrtnmgs @sshine That commit contains code to align line numbers for each font-size in the new responsive design

sshine:
OK, so a proper solution does not remove this code but finds a way to make the line numbers align at multiple font sizes? The new responsive design is the one that was rolled out when the site became mobile friendly, right?

PR #496, opened 2019-01-11
rename "Mentor this solution" to "Review this solution"

AndreaCrotti:
I find "Mentor this solution" quite weird since you normally "Mentor" a person, not a solution.
So I just replaced it with "Review", and changed the tests as well.

Haven't set it up all locally yet so I'm not 100% sure it's correct, but I think that was the only place to change from some ack magic.

### DIFF
--- a/app/views/mentor/solutions/_claim-section.html.haml
+++ b/app/views/mentor/solutions/_claim-section.html.haml
@@ -11,7 +11,7 @@
   .request
     %h3 Give feedback on this solution
     %p Can you help this learner improve this solution? Do you feel it's ready to approve? Click the button below to start giving feedback.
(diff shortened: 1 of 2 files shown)


iHiD:
@AndreaCrotti This is a nice improvement. Thank you 💙

F3PiX:
I agree that we normally use 'mentoring' for people, not for code, but I find it more important to communicate that Exercism mentoring is not exactly the same as a code review.  I'd vote for either leaving it as it is, or find an alternative that takes the difference with a normal code review into account.

iHiD:
@F3PiX We cross-posted there. Sorry. I've deployed this now. Happy to discuss this further. Maybe you could open one on exercism/exercism to get a wider set of suggestions?

PR #498, opened 2019-01-15
Add button to unpublish solution on 'my solution' page

PatrickMcSweeny:
Closes exercism/exercism#4525. I was motivated into action by the continued discussion on exercism/exercism#4163.

Side note: I unexpectedly ran into some trouble with the latest_exercise_version? method while writing the system tests. I wonder if that is an issue that needs to be addressed to make sure it doesn't problems for other people writing tests.


PatrickMcSweeny:
If this section of the UI is being redesigned, is this pull request unnecessary?

iHiD:
Probably. We still need to think through the consequences of unpublishing solutions as part of things.

PatrickMcSweeny:
I'm confused. It is currently possible to unpublish a solution from the profile page, this was just supposed to create a new UI for that.

PR #514, opened 2019-01-21
Thicken font weights

iHiD:


yyyc514:
I thought it was bigger not bolder, but it's definitely a legibility improvement - and I personally don't have any particular accessibility concerns.

PR #517, opened 2019-01-22
close exercism/exercism#4602

Smarticles101:
closes exercism/exercism#4602


Smarticles101:
There you go @iHiD thanks for catching that, I wasn't really thinking about the commit message at the time

iHiD:
Great. Thanks! :)

PR #533, opened 2019-01-27
Fix comment ordering

iHiD:


iHiD:
Good catch. Fixed by https://github.com/exercism/website/pull/540 I should have added tests!

PR #535, opened 2019-01-29
Add some respect for Octokit::TooManyRequests

iHiD:


app/services/git/sync_updated_repos.rb
```diff
@@ -25,6 +25,9 @@ def sync
     repo_updates.each do |update|
       begin
```
F3PiX: Could you use `find_each`, and then maybe set a batch size? I worked on a project where we did that to prevent TooManyRequest errors in the first place.

PR #538, opened 2019-01-29
Fix navbar

iHiD:
(cc @PatrickMcSweeny)


PatrickMcSweeny:
It appears that the image url might be part of the problem. This doesn't load anything:

https://assets.exercism.io/assets/nav-divider.png

iHiD:
Assets are served from s3 in production. `image-url` is contextual. It will behave differently in production to development as it respects the asset location config setting. `url` will just load a relative URL, which works in dev, but then will fail in production.

PR #578, opened 2019-02-24
Make widget title the same height as selectbox

EarthlingRich:
The alignment of the selectbox widget looks a bit broken to me. The title has a different height than te selectbox. This PR makes it the same height.

**Old**
afbeelding

**New**
afbeelding


iHiD:
@EarthlingRich Thanks. This already looks fine on master to me (see screenshot below). What OS/browser combination are you using?

screenshot 2019-02-26 at 12 35 02

PR #600, opened 2019-03-18
Cybot1711 patch 1

cybot1711:
Render Crips fonts.


iHiD:
Hello. Thanks for this. Could you explain why this is preferred to the defaults please? As I don't have the capacity to test this on multiple devices I need to understand the effect this has on different browsers on different OSs, why the browser defaults are wrong, and what the side effects might be. I note GitHub doesn't do it (which is what I tend to base our font rendering decisions on) so I'd be interested to learn about why that is. Thank you! :)

cybot1711:
> Hello. Thanks for this. Could you explain why this is preferred to the defaults, please? As I don't have the capacity to test this on multiple devices I need to understand the effect this has on different browsers on different OSs, why the browser defaults are wrong, and what the side effects might be. I note GitHub doesn't do it (which is what I tend to base our font rendering decisions on) so I'd be interested to learn about why that is. Thank you! :)

@iHiD 
I would not say that it is preferred according to the standard. This change will have absolutely no side effects other than not rendering crisp fonts where it is not supported. [Check this](https://medium.com/@MateMarschalko/improving-font-rendering-with-css-3383fc358cbc)

iHiD:
Hey @cybot1711 - just so you know, I've decided *not* to accept this PR yet, but I just need some actual time to think about it and test it first, and there are bigger priorities I'm dealing with atm. In case you wanted to keep it open :)

PR #618, opened 2019-04-06
Fix test that breaks after spring hour change

ccouzens:
In England, Daylight saving time ended 31st of March. This meant the
`time_ago_in_words calculates time in days, not months` test was
returning 32 days for me, not 33 days.

Here's how it worked.
If I ran the test at 2019-04-06 22:00.
Time.now = 2019-04-06 22:00.
33.days = ActiveSupport::Duration object of 33 days
Time.now - 33.days = 2019-03-06 22:00.
2019-04-06 22:00 - 2019-03-06 22:00 = 32 days 23 hours.

I fix it by having the test rewind in a manner which doesn't take into
account time jumps.

I think it is the test that is wrong, not the code.
If someone did something 32 days and 23 hours ago, it wouldn't be
accurate to say it was 33 days ago. Although possibly we could round up
if it's nearly a full day.

### DIFF
--- a/test/helpers/time_ago_in_words_helper_test.rb
+++ b/test/helpers/time_ago_in_words_helper_test.rb
@@ -2,15 +2,15 @@ require "test_helper"
 
 class TimeAgoInWordsHelperTest < ActionView::TestCase
   test "time_ago_in_words calculates time in days, not months" do
-    assert_equal "33 days", time_ago_in_words(Time.now - 33.days)
+    assert_equal "33 days", time_ago_in_words(33.days.ago)
   end
 
   test "time_ago_in_words calculates time in days" do
-    assert_equal "2 days", time_ago_in_words(Time.now - 2.days)
+    assert_equal "2 days", time_ago_in_words(2.days.ago)
   end
 
   test "time_ago_in_words calculates time less than one day in hours or minutes" do
-    assert_equal "about 2 hours", time_ago_in_words(Time.now - 2.hours)
+    assert_equal "about 2 hours", time_ago_in_words(2.hours.ago)
   end
 
   test "time_ago_in_words correctly calculates one day plus one second" do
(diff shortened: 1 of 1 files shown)


iHiD:
Hey. Thank you for catching this and for writing the PR and the explanation :)

I'm marginally unsure how (and slightly whether) the code is correct. My confusion is this:

```
2.5.1 :002 > (Time.now - 33.days).to_i
 => 1551742300 
2.5.1 :003 > (Time.now - 33.days.to_i).to_i
 => 1551738710 
```

You said
> I fix it by having the test rewind in a manner which doesn't take into account time jumps.

Could you explain how the inner `to_i` does that please?

An alternative solution, which might feel a little less esoteric, might be to change the `Time.now` to `Time.current`, which is the standard we use everywhere else. 

Thoughts?

ccouzens:
Hi @iHiD,

> Could you explain how the inner `to_i` does that please?

`ActiveSupport::Duration`'s subtraction seems to work by looking at the expanded form of the date. Having specified that we want to go back 33 days, it seems to be wanting to keep the time component the same. I don't know if this is a bug or a feature.

```ruby
irb(main):028:0> t = Time.now; puts t; puts t - 33.days
2019-04-07 12:11:07 +0100
2019-03-05 12:11:07 +0000
```

The `to_i` converts 33 days into seconds. This then correctly incorporates the hour change.

```ruby
irb(main):029:0> t = Time.now; puts t; puts t - 33.days.to_i
2019-04-07 12:14:14 +0100
2019-03-05 11:14:14 +0000
```


> An alternative solution, which might feel a little less esoteric, might be to change the Time.now to Time.current, which is the standard we use everywhere else.

I'm concerned that might be brittle to an unrelated change.

```ruby
irb(main):022:0> Exercism::Application.config.time_zone
=> "UTC"
irb(main):020:0> t = Time.current; (t - (t - 33.days)).to_i / 24.0/60/60
=> 33.0
irb(main):021:0> t = Time.find_zone('America/New_York').now; (t - (t - 33.days)).to_i / 24.0/60/60
=> 32.958333333333336
```

It works because Rails defaults to UTC and UTC doesn't have hour changes. It's unlikely exercism would be reconfigured to be a different time zone, but if it was, we wouldn't want this test to break.

----

```ruby
33.days.ago
```

seems to be performing correctly in all instances. Perhaps that would be the best solution?

ccouzens:
I've updated to be easier than it was at the start.

PR #622, opened 2019-04-10
Fix issues arising from renaming exercises

iHiD:
Closes https://github.com/exercism/javascript/issues/664

### DIFF
--- a/app/models/api/solution_responder.rb
+++ b/app/models/api/solution_responder.rb
@@ -98,7 +98,7 @@ class API::SolutionResponder
   private
 
   def exercise_reader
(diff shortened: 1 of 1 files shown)


SleeplessByte:
@iHiD mind elaborating on _what_ exactly went wrong. Why are these two not in-sync?

iHiD:
Solutions are locked to the git sha that the person downloads (so that the exercise doesn't change under their feet while they're solving it). We retrieve the exercise details at the stored sha of the repo, and in the old sha that we were retrieving, the filepath in the repo was the old slug. So we need to honour that for things to work properly.

PR #631, opened 2019-04-25
Fix build to work on Travis `xenial`

ccare:
The default image for Travis has changed. This commit fixes the build
to reflect that. Specifically, there is now no need to upgrade mysql
and no need to manage xvfb ourselves.


iHiD:
Brilliant. Thanks, @ccare.

PR #664, opened 2019-07-01
Bind scroll events after a certain timeout

kntsoriano:
Closes https://github.com/exercism/bugs/issues/1.

Since we don't have direct control on knowing when the walkthrough fully loads, we just need to bind the scroll events after a certain timeout. In the future, we can definitely refactor for the walkthrough to fire an event when it loads.


iHiD:
Nice work :)

PR #665, opened 2019-07-02
Don't delete secrets.yml during deployment

kntsoriano:
Since we've removed `secrets.yml` from version control, we should also delete the step where we delete it during deployment. This is blocking me to deploy so I hope you can take a look at this as soon as possible.


kntsoriano:
@iHiD This is the commit 6a9ae18322fc370ad0b12346bb4556fd18b7c917 that does the symlink.

iHiD:
OK. I think this can be merged. I've pushed a commit to add `config/secrets.yml` to `.gitignore` as I nearly committed it a couple of times :S

PR #671, opened 2019-07-03
Cache bundler on CI

kntsoriano:
## Description
As I was going through the PRs, I noticed that we take too much time running the tests on CI and this further impacts us when I have PRs piled up that I want to merge. The two longest processes are:

1. Running `bundle install`.
2. Running system tests.

This PR fixes (1) by caching the bundler produced files.

### Before (10 mins to run tests)
Screen Shot 2019-07-03 at 10 55 04 AM

### After (5 mins to run tests)
Screen Shot 2019-07-03 at 11 03 29 AM

## Other information
1. In order to enable caching, I also needed to turn on the "Build pushed branches" setting as outlined in the Travis CI docs. As we've turned this on, we need to modify `.travis.yml` to only build the master branch when pushed (PRs are an exception, we still build them). This will allows us to use the master branch's cache for subsequent PRs to master.
2. I also opted to turn on auto cancellation of builds in order for Travis CI to only build the latest commits in the branch. This will help us free up workers if we push frequently to branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ cache:
   node_modules: true
   directories:
     - vendor/bundle
-
+branches:
+  only:
+  - master
 after_script:
   - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT -t simplecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
   - sudo mysql -e "ALTER DATABASE exercism_reboot_test CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;" -u root
   - sudo mysql -e "GRANT ALL PRIVILEGES ON exercism_reboot_test.* TO 'exercism_reboot'@'localhost'" -u root
 install:
-  - "bundle install"
+  - "bundle install --jobs=3 --retry=3 --deployment"
   - "npm install"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ before_install:
   - sudo mysql -e "ALTER DATABASE exercism_reboot_test CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;" -u root
   - sudo mysql -e "GRANT ALL PRIVILEGES ON exercism_reboot_test.* TO 'exercism_reboot'@'localhost'" -u root
 install:
-  - "bundle install"
+  - "bundle install --jobs=3 --retry=3 --deployment"
```
kntsoriano: This command is what Travis CI uses in order to install dependencies. It uses parallel jobs to make it faster, and also adds the `--deployment` flag in order to write the gems into `vendor/bundle`.

As we see below in this file, we cache `vendor/bundle` but we aren't able to cache anything because we left out the `--deployment` flag.

iHiD:
Great work :) Thanks dude!

iHiD:
And thank you for explaining it all so clearly :)

iHiD:
@kntsoriano Can we speed up PRs too? It's still taking 5mins just to build the gems for PRs.

PR #686, opened 2019-07-24
Add soft deletion for user

kntsoriano:
Closes https://github.com/exercism/bugs/issues/16.

According to the [Bugsnag report](https://app.bugsnag.com/exercism/exercism/errors/5c4de88587ed6900193413f4), users are unable to delete their accounts once they have approved a solution. As we want to keep foreign key integrity, I decided to implement a "soft-delete" functionality for our user model. Let me know what you think of this.

### DIFF
--- a/app/services/user_services/delete.rb
+++ b/app/services/user_services/delete.rb
@@ -2,14 +2,19 @@ module UserServices
   class Delete
     include Mandate
 
-    initialize_with :user
+    def initialize(user, time: Time.current)
+      @user = user
+      @time = time
+    end
 
     def call
       redact_user_details
       delete_user_associations
+      mark_as_deleted
     end
 
     private
+    attr_reader :user, :time
 
     def redact_user_details
       user.skip_reconfirmation!
@@ -61,6 +66,10 @@ module UserServices
       user.email_log.destroy if user.email_log
       user.profile.destroy if user.profile
       user.communication_preferences.destroy if user.communication_preferences
+    end
+
+    def mark_as_deleted
+      user.update!(deleted_at: time)
     end
   end
 end
--- a/app/services/user_services/delete.rb
+++ b/app/services/user_services/delete.rb
@@ -36,8 +36,6 @@ module UserServices
         provider: nil,
         uid: nil,
         admin: false,
-        accepted_privacy_policy_at: nil,
-        accepted_terms_at: nil,
         dark_code_theme: false,
         is_mentor: false,
         default_allow_comments: nil,
--- a/app/services/user_services/delete.rb
+++ b/app/services/user_services/delete.rb
@@ -1,0 +1,41 @@
+module UserServices
+  class Delete
+    include Mandate
+
+    initialize_with :user
+
+    def call
+      redact_user_details
+    end
+
+    private
+
+    def redact_user_details
+      user.skip_reconfirmation!
+      user.update!(
+        name: "Deleted User",
+        handle: "deleteduser#{SecureRandom.uuid}",
+        bio: nil,
+        email: "deleteduser+#{SecureRandom.uuid}@exercism.io",
+        remember_created_at: nil,
+        sign_in_count: 0,
+        current_sign_in_at: nil,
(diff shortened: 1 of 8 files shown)


db/migrate/20190724042815_add_deleted_at_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeletedAtToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :deleted_at, :timestamp
```
iHiD: Can we make this a `datetime`, not a `timestamp` pls? I prefer to store dates as dates.

ccare:
This is really interesting. So long as @kntsoriano  is happy that we're blanking our all Personal information, I think this is fine.

To understand the user experience I had a look at the form and I think it's really clear that it's  irreversable and also clear that we we're not removing comments, so that's good.

I have a few thoughts though

1) Are we happy about blanking out the IP addresses? There's a potential scenario where bots could sign up, post a load of malicious comments and then de-register, leaving very little evidential trace. In particular we would have blanked out these:

        current_sign_in_ip: nil,
        last_sign_in_ip: nil,

We need to make sure that we still have enough info for blocking bad bots etc. Maybe this is a job for operational logs though, rather than the database.

2) Should we retain the information about when the user confirmed our legals? These are these are the relevant fields:

     accepted_privacy_policy_at: nil,
     accepted_terms_at: nil,

It could be very handy to know that a user signed up under a given version of the terms (it might impact how we use the anonymised data in the future).

3) Not strictly related to this: Should we be checking that the checkbox 'I understand that this is irreversible' has actually been checked? It's currently managed by HTML validation, but I wonder if a server-side guard would be a safer check (given the irreversible nature of the action).

4) Should we, somewhere, mention that we have database backups, so that there will be some 'residual' data held for a period of time. I.e. that user's information may be held in operational systems for a period of time before their data is entirely erased.

5) Linked to (4) - if we were to ever fallback to a historic backup, how would we ensure that we re-applied any user-deletions to the restored version? Do we want to log the deletion request somewhere else as well?

app/services/user_services/delete.rb
```diff
@@ -0,0 +1,75 @@
...
+        current_sign_in_ip: nil,
+        last_sign_in_ip: nil,
+        confirmation_token: nil,
+        confirmed_at: nil,
+        confirmation_sent_at: nil,
+        unconfirmed_email: nil,
+        provider: nil,
+        uid: nil,
+        admin: false,
+        accepted_privacy_policy_at: nil,
```
iHiD: Remove this line ^^

app/services/user_services/delete.rb
```diff
@@ -0,0 +1,75 @@
...
+        accepted_terms_at: nil,
```
iHiD: Remove this line ^^

iHiD:
@ccare Great comments. Thanks. My immediate thoughts.

1) I'm fine with deleting this data. We log the user id in the logs alongside IP address, so we have that data there. Plus IP addresses are PII so we probably *have* to delete them in this context?
2) Yes. Good point. @kntsoriano - I've added inline comments for this.
3) Let's not block this PR on this, but I agree. @kntsoriano I've added https://github.com/exercism/wip/issues/23 for this
4) I'd like to put this in the T&Cs rather than putting it here. (When you request your data is deleted,  we will do XYZ). I think that's good as a thing to have for clarity anyway.  I've added an issue for you [here](https://github.com/exercism/wip/issues/24), @ccare.
5) As a short-term solution, we could parse your new shiny logs for requests to that endpoint. But having an extra "events" log that we can write things like this too might be really good generally. I like that idea a lot. Do you want to think about that and make an issue for it if appropriate, @ccare ?

db/migrate/20190724042815_add_deleted_at_to_user.rb
kntsoriano: `timestamp` and `datetime` in Rails is the same thing.

PR #687, opened 2019-07-24
Add some words about "download in mentor mode"

glennj:
I'm a fairly new mentor, and I didn't know that "mentor mode download" would not overwrite my own solution to the exercise.


glennj:
No longer relevant

iHiD:
You are also no longer a fairly new mentor! :D

PR #693, opened 2019-08-08
Add system message when a user requests feedback for an auto approved solution

kntsoriano:
Closes https://github.com/exercism/wip/issues/25.

Is there a difference between an exercise being marked as "auto approved" and "approved by system"?


iHiD:
\rebase

iHiD:
@kntsoriano I'll look at this soon for you. It slipped past my radar. Sorry.

kntsoriano:
@iHiD Just fixed the failing tests. Should be ready for review.

kntsoriano:
@iHiD You might've missed this.

app/models/system_discussion_post.rb
```diff
@@ -0,0 +1,13 @@
+class SystemDiscussionPost
```
iHiD: Can you change this to be a service (`CreateSystemDiscussionPost`) like the other ones that already exist pls (although it doesn't have to inherit etc like the others).

app/services/request_mentoring_on_solution.rb
```diff
@@ -9,6 +9,26 @@ def call
       return if user_track.mentoring_allowance_used_up?
     end
 
+    post_system_messages!
```
iHiD: I think I'd prefer this to be `post_system_messages! if solution.approved_by_system?` as it's clearer when reading the call method as to why we're creating system messages (and remove the guard below). If you disagree, I don't mind you leaving it as it is though

test/system/my/request_mentor_feedback_test.rb
```diff
@@ -0,0 +1,17 @@
+require 'application_system_test_case'
```
iHiD: What do you think about renaming this file to `test/system/request_mentoring_on_solution_test.rb` so it's clearer that it's linked to the service if we go looking in the future? I'm always unsure how to name system tests. We should probably agree a standard.

test/system/my/request_mentor_feedback_test.rb
```diff
@@ -0,0 +1,17 @@
...
+
```
iHiD: What do you think about adding a second test for the case where it's the notification shouldn't get added, to guard against a future accidental regression?

app/services/request_mentoring_on_solution.rb
kntsoriano: Cool. Sounds good.

test/system/my/request_mentor_feedback_test.rb
kntsoriano: I think we shouldn't have any linking with what the system test is named and what the service is named as system tests are high level tests. I think with regarding to naming system tests, we just need to make sure that the file name is understandable as to what it pertains to. We can move to naming system tests as user stories such as, `user_requests_mentor_feedback_test.rb`.

test/system/my/request_mentor_feedback_test.rb
kntsoriano: I will add this to the service level test. Thanks for reminding me.

test/system/my/request_mentor_feedback_test.rb
iHiD: Yeah, I like the story idea.

PR #694, opened 2019-08-08
Upgrade prism JS

kntsoriano:
Closes https://github.com/exercism/bugs/issues/17.

Upgrading prism to the latest version fixes this error. I also tried upgrading the CSS in order for it to be with sync in JS but I'm having trouble to do so. Have we added anything extra? In particular, I'm having difficulty keeping the default theme separate from the dark theme.

Is there also any reason for us not to be using `yarn` for this dependency?


iHiD:
\rebase

kntsoriano:
I'll be opening a new PR for this.

PR #696, opened 2019-08-13
Update font for wider native support

iHiD:
This adds the new `system-ui` property, but leaves `-apple-system, BlinkMacSystemFont`, which it replaces, because browser support for `system-ui` isn't wide enough yet. It also adds in some extra fonts for Linux (generally), Ubuntu and Gnome.

It should make no visible difference on Mac.


iHiD:
\rebase

PR #703, opened 2019-08-26
Upgrade Prism JS

kntsoriano:
Closes https://github.com/exercism/bugs/issues/17.

### PrismJS installation
The PrismJS docs states three ways to install it: (1) downloading the source from their website and hosting it, (2) using a CDN, (3) same as (1) but using a JS bundler such as webpack or browserify.

As we agreed earlier, we wanted to do (3), but I was unable to do so as our codebase at the moment doesn't use webpacker. I've settled into using (2) because of the ff reasons:

1. (1) makes it hard for us to upgrade and will require me to write a lengthy documentation describing the steps a future developer needs to do in order to upgrade Prism. It isn't straight forward as downloading the source as we need to customize Prism to include plugins we are using. While I was doing the feature, this took a bit of time for me to figure out.
2. (3) is the optimal solution, but right now, it might be too risky to introduce webpacker into this codebase, it'll be hard for me to ensure that I don't break anything.

We need to decide if the cons of having Prism hosted on a CDN outweigh what I've outlined above.

### Removal of prism-clojure
Going through our commit history, I figured that `prism-clojure` was included as Prism didn't include it by default. But in the current version of Prism that we're using, it is already included.

### Toggling of themes
I've decided to change how we toggle themes by swapping out the stylesheet in `<head>`, this is how the prism team does swapping of themes on their website as well. https://github.com/PrismJS/prism/issues/643.

#### Original
Screen Shot 2019-08-26 at 6 02 41 PM

#### After upgrade
Screen Shot 2019-08-26 at 6 02 12 PM

In the light theme, code blocks in discussions had their colors overriden to look like a normal `<pre>` tag. I assume this wasn't an intended feature and thus I've done https://github.com/exercism/website/pull/703/commits/5630c2661ded9eee6ed41bf1e512c3b78d7b73ab. Please let me know if it was otherwise.

## Testing
Here's a screencast of me going through the pages, and checking I've changed anything diff wise, please let me know if I've missed anything.

https://www.loom.com/share/182e01cd766547b791d1c502d80f1566


app/assets/stylesheets/code-editor.scss
```diff
@@ -54,9 +54,19 @@ body .md-editor {
         text-decoration: underline;
     }
 
-    pre {
+    pre:not([class*="language-"]) {
```
iHiD: > In the light theme, code blocks in discussions had their colors overriden to look like a normal `<pre>` tag. I assume this wasn't an intended feature and thus I've done 5630c26. Please let me know if it was otherwise.

That was intentional, so please reverse this one.

app/assets/stylesheets/prism-modifications.scss
```diff
@@ -4,6 +4,7 @@
 pre.line-numbers {
     margin: 0 !important;
     padding-bottom: 1em !important;
+    border-radius: 0 !important;
```
iHiD: What's the reason for this change? I couldn't see it from the video and I can't really see the change when I compare branches)

kntsoriano: There's a 0.3em border-radius when you inspect the CSS. Not really visible, but decided to remove it as it might create some unexpected behaviors.

app/assets/stylesheets/code-editor.scss
kntsoriano: Thanks.

iHiD:
Great. Thanks dude :)

PR #704, opened 2019-08-28
Add recaptcha to sign up page

kntsoriano:
Closes https://github.com/exercism/wip/issues/28.

Screencast: https://www.loom.com/share/bc5f474a84c5477392dfb86f93c1e0d2

Issue in screencast: https://github.com/google/recaptcha/issues/248

### DIFF
--- a/app/assets/stylesheets/widgets/recaptcha.scss
+++ b/app/assets/stylesheets/widgets/recaptcha.scss
@@ -1,0 +1,5 @@
+.g-recaptcha {
+  > div {
+    margin: auto;
+  }
+}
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/widgets/recaptcha.scss
```diff
@@ -0,0 +1,5 @@
+.g-recaptcha {
+  > div {
+    margin: auto;
+  }
+}
```
kntsoriano: This is to center the recaptcha box. We have no way of making it's width longer, thus centering might be our best bet. Happy to remove this otherwise.

config/recaptcha.yml
```diff
@@ -0,0 +1,5 @@
+development:
```
kntsoriano: These are keys to the development version of our Recaptcha. I will regenerate keys for production and will insert this yml file during deploy.

kntsoriano:
Deployed and made sure that user sign up still works.

iHiD:
Great. Thank you 💙 

I'll check on Monday to ensure we've not seen a drop in the number of signups we'd expect.

PR #711, opened 2019-09-04
Prevent users who haven't submitted a solution to sign up as mentors

kntsoriano:
https://github.com/exercism/wip/issues/33

Screenshot:
Screen Shot 2019-09-04 at 1 45 58 PM


F3PiX:
This is awesome! 
Is this checking for solutions in the track they are going to mentor? Or any solutions?
(It's awesome either way, but even more when it's for each track.)

ErikSchierboom:
Great idea!

iHiD:
The main purpose for now is just to stop people from being able to join and spam people (we're getting spam bots).

SleeplessByte:
This might be to reject spam bots, but I actually really like this idea in general! Preventing people from mentoring exercises they haven't done!

kytrinyx:
There are a lot of people who are very accomplished in languages who have no interest in doing the exercises themselves, so in general I wouldn't want them to have to have done _all_ the exercises, but to have tried at least something in the track seems perfectly reasonable.

app/controllers/mentor/registrations_controller.rb
```diff
@@ -6,12 +6,23 @@ def index
...
+    load_tracks
+    @errors = []
   end
 
   def create
-    MakeUserAMentor.(current_user, params[:track_id])
-    redirect_to mentor_dashboard_path
+    command = MakeUserAMentor.new(current_user, params[:track_id])
+
+    command.call
```
iHiD: Maybe `command.()` for consistency's sake (I figure it'll be easier to search for when we add this monad style thing into the command patter on mandate).

PR #695, opened 2019-08-08
Add rel='nofollow' to new links in markdown to discourage spam

mjc:
Currently user-generated links count as helping the SEO rank of destination websites.  Adding rel="nofollow" should reduce the incentive for spam commenters.  I can add a test but I wanted to float this as an idea, first.

PR #714, opened 2019-09-09
Add optional rel='nofollow' to links

iHiD:
Currently user-generated links count as helping the SEO rank of destination websites. Adding rel="nofollow" should reduce the incentive for spam commenters.

This replaces #695.


iHiD:
@mjc I'm merging this now. Thank you.

PR #715, opened 2019-09-10
Make testing easier

kntsoriano:
Closes https://github.com/exercism/wip/issues/39.

Instead of introducing commands, I decided that it would be collectively easy for us if anything that has to do with Git repos are stubbed by default in the test environment.


app/models/git/problem_specifications.rb
```diff
@@ -1,8 +1,10 @@
 class Git::ProblemSpecifications < Git::RepoBase
-  REPO_URL="https://github.com/exercism/problem-specifications"
+  class << self
```
iHiD: Rather than using this syntax, could we use `def.self` and `cattr_reader`. I prefer defining methods on the class rather than on the eigenclass as a general rule. If there's a reason you're doing it this way that I've not understood, let me know :)

kntsoriano: Using `cattr_accessor` creates a class variable and not a class instance variable. When using class variables, it gets tricky when dealing with inheritance as class variables when changed in the sub class, changes the class variable in the parent class.

iHiD: @kntsoriano That makes sense. Thanks for the explanation :)

PR #721, opened 2019-09-16
Add flipper for feature flags

kntsoriano:
In preparation for creating the changelog, I've decided to add in feature flags in order to make small PRs to master possible. This will allow reviewers to easily track changes and to follow features as they're being developed.

Screencast: https://www.loom.com/share/0705d01d44b14b15a0594f14ab425992

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -1,0 +1,15 @@
+Flipper.configure do |config|
+  config.default do
+    adapter = if Rails.env.test?
+                Flipper::Adapters::Memory.new
+              else
+                Flipper::Adapters::ActiveRecord.new
+              end
+
+    Flipper.new(adapter)
+  end
+end
+
+Flipper.register(:admins) do |actor|
+  actor.respond_to?(:admin?) && actor.admin?
+end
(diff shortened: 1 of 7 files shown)


config/initializers/flipper.rb
```diff
@@ -0,0 +1,15 @@
...
+              else
+                Flipper::Adapters::ActiveRecord.new
+              end
+
+    Flipper.new(adapter)
+  end
+end
+
+Flipper.register(:admins) do |actor|
+  actor.respond_to?(:admin?) && actor.admin?
```
iHiD: I think this be reduced down to using try?

```suggestion
  actor.try(:admin?)
```

kytrinyx: Or even `actor&.admin?`

iHiD: Does `&` defend against a non-nil object without the method defined?

IIRC:
- `"foo".try(:non_existant_method)` returns `nil`
- `"foo".try(&:non_existant_method)` raises an exception.

I don't know which way `&` works and am too tired to Google :)

kntsoriano: I just copy pasted this from the wiki, actually. I will remove all `respond_to?` calls as we're dealing with the User model here.

PR #722, opened 2019-09-19
Testing submissions

iHiD:
This adds a prototype for testing submissions.


app/views/my/solutions/solve.html.haml
```diff
@@ -0,0 +1,105 @@
...
+          method: "PATCH",
+          data: {
+            files: {
+              "two_fer.rb": code
+            }
+          }
+        }
+      ).done(function(data) {
+        submissionId = data.submission.id
+        $('#tests-output .ops-status').show().children('span').html("Tests pending")
```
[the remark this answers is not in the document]
SleeplessByte: When you see the following pattern:

```javascript
$("parent child")
$("parent child")
$("parent child")
```

You probably should memoize it, so that when you _change_ the selector, you don't need Shotgun Surgery:

```javascript
var $opsStatus = $("#tests-output .ops-status")
$opsStatus.show().children('span')...
```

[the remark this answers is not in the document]
SleeplessByte: When you see the following pattern:

```javascript
$("parent child1")
$("parent child2")
$("parent child3")
```

it can make sense to extract parent first. This automatically indicates a (sub)-component:

```javascript
var $testsOutput = $('#tests-output')

var $opsStatus = $testsOutput.find('.ops-status')
var $opsStatus = $testsOutput.find('.results-status')
var $opsStatus = $testsOutput.find('.results-message')
```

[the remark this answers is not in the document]
SleeplessByte: Right now you're using `classes` to determine where things go. That's _okay_, but not great. Classes can be re-used, and you're mixing IDS with classes, but you _always_ only select the first result from the query. This means that if I add a second `.ops-status` on the page, it might work, breaking the old one, or might not work, preservering the old one.

This probably indicates that you want scoping. The scope is not an HTML scope, but a "data" scope. It's the ops-status of the tests you're running. So the scope is probably "the current iteration of the current exercise in the current track".

app/views/my/solutions/solve.html.haml
```diff
@@ -0,0 +1,105 @@
...
+              $('#tests-output .ops-status').show().children('span').html("Tests finished")
+              $('#tests-output .results-status').show().children('span').html(data.status)
+
+              if(data.message) {
+                $('#tests-output .results-message').show().children('span').html(data.message)
+              }
+
+              if(data.failure) {
+                $('#tests-output .test-failure').show()
+                $('#tests-output .test-failure .name').html(data.failure.name)
```
SleeplessByte: Are you _certain_ you want HTML here? It's probably returned as text, so use `.text` instead. `.html` can:
- break things when you don't escape the HTML
- is an entry for XSS

app/views/my/solutions/solve.html.haml
```diff
@@ -0,0 +1,105 @@
...
+            }
+          }
+        }
+      ).done(function(data) {
+        submissionId = data.submission.id
+        $('#tests-output .ops-status').show().children('span').html("Tests pending")
+
+        for(sid in intervals) { clearInterval(intervals[sid]) }
+
+        intervals[submissionId] = setInterval(function() {
```
SleeplessByte: This does NOT what you want it to do. I'll explain:

If I understand correctly you want to:
- submit the code
- show pending
- each 500 ms check if it's done by getting the test results
- if it is, stop the interval and be done

But the way this is written is not optimal. It's:
- relying on the the request to _fail_ when it's not done
- has various race-conditions.

Consider the following. Let's say it takes 1.1s for test_results to return. It takes 2 seconds for the tests to run

- t: 0 .done code is executed
- t: 500: ➡ first ajax request to get results is _started_ 
- t: 1000: ➡ second ajax request to get results is _started_
- t: 1500: ➡ third ajax request to get results is _started_
- t: 1600: 🔙 first ajax request is _received_, this is an error, ❌ throw away
- t: 2000: ➡ fourth ajax request to get results is _started_
- t: 2100: 🔙 second ajax request is _received_, this is an error, ❌ throw away
- t: 2500: ➡ fifth ajax request to get results is _started
- t: 2600: 🔙 third ajax request  is _received_, this MIGHT have the result. Let's say it does. ✔ executes callback, no new ajax requests will be scheduled
- t: 3100: 🔙 fourth ajax request  is _received_, with result ✔ executes callback
- t: 3600: 🔙 fifth ajax request  is _received_, with result ✔ executes callback

A quick fix, to stop the "double renders/executes" is to, at line 87, break out of the function:

```javascript
              if(intervals[submissionId]) { 
                clearInterval(intervals[submissionId])
                delete intervals[submissionId]
              } else {
                return /* double render */
              }
```

... but better would be to use `setTimeout` instead, and `onError`, start a new request. Yes, it will take _up to a roundtrip_ longer to see the results, but it will save you so many requests:

- t: 0 .done code is executed
- t: 500: ➡ first ajax request to get results is _started_ 
- t: 1600: 🔙 first ajax request is _received_, this is an error, ❌ throw away
- t: 1600: ➡ error handler for ajax starts new request 
- t: 2700: 🔙 second ajax request is _received_, this MIGHT have the result. Let's say it does. ✔ executes callback, no new ajax requests will be scheduled

If calls come back _within 500 ms, you can also choose to schedule the next one _a but further out_:

```javascript
whenError((response) => {
  const timeSinceLastRequest = convertToUnix(response['Date']) - previousRequestTime
  intervals[submissionId] = setTimeout(requestResults, Math.max(500 - timeSinceLastRequest, 0))
})
```

app/views/my/solutions/solve.html.haml
```diff
@@ -0,0 +1,105 @@
...
+              $('#tests-output .results-status').show().children('span').html(data.status)
+
+              if(data.message) {
+                $('#tests-output .results-message').show().children('span').html(data.message)
+              }
+
+              if(data.failure) {
+                $('#tests-output .test-failure').show()
+                $('#tests-output .test-failure .name').html(data.failure.name)
+                $('#tests-output .test-failure .message').html(data.failure.message.replace("\n", "<br/>"))
```
SleeplessByte: Are you _certain_ you want HTML here? It's probably returned as text, so use `.text` instead. `.html` can:
- break things when you don't escape the HTML
- is an entry for XSS

I see that you're manually building new lines. Is this a "list of errors", or do you want to "keep the original format".
- in case of a list, use a `ul` or `ol` instead, and build `li`s instead of doing this
- in the other case, that is exactly what `pre` is for.

app/views/my/solutions/solve.html.haml
```diff
@@ -0,0 +1,105 @@
...
+    .results-message,
+    .test-failure {
+      display:none;
+    }
+  }
+  #run-tests-button {
+    background:green;
+  }
+
+-content_for :js do
```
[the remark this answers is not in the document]
iHiD: Yeah, totally. They would go in proper JS files now as well. I'm just trying not to pollute the main asset pipeline with this prototype, so that I don't have to think about it too much. Same for the `sass`, which IMO should never be compiled in the HAML/ERB as its very slow.

app/views/my/solutions/solve.html.haml
iHiD: Didn't know that `.text` existed. Thanks.

app/views/my/solutions/solve.html.haml
iHiD: > in the other case, that is exactly what pre is for.

Clever @SleeplessByte!

iHiD:
@SleeplessByte Thank you!! 💙💜❤️💛💚💖

app/views/my/solutions/solve.html.haml
SleeplessByte: The non-jQuery equivalent is `innerText`, but I recommend [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent), as it can prevent XSS and doesn't "hide" hidden text, and doesn't have the 'destroy everything "bug"' in IE11.

Then again, using jQuery is fine.

PR #726, opened 2019-09-22
list most recent testimonials first

yyyc514:
- applies a sort order to the `solution_mentorships` query for
  loading testimonials for the Dashboard

### DIFF
--- a/app/controllers/mentor/dashboard_controller.rb
+++ b/app/controllers/mentor/dashboard_controller.rb
@@ -101,6 +101,7 @@ class Mentor::DashboardController < MentorController
     @testimonials = current_user.solution_mentorships.
                         with_feedback.
                         includes(solution: [:user, {exercise: :track}]).
+                        order('solutions.completed_at DESC').
                         page(params[:page]).per(20)
   end
 
--- a/app/controllers/mentor/dashboard_controller.rb
+++ b/app/controllers/mentor/dashboard_controller.rb
@@ -71,7 +71,10 @@ class Mentor::DashboardController < MentorController
     @total_solutions_count = current_user.solution_mentorships.count
     @total_students_count = current_user.mentored_solutions.select(:user_id).distinct.count
     @weekly_solutions_count = current_user.solution_mentorships.where("solution_mentorships.created_at > ?", Time.now.beginning_of_week).count
-    @feedbacks = current_user.solution_mentorships.where(show_feedback_to_mentor: true).order('updated_at desc').limit(5).pluck(:feedback)
+    # TODO: re-reenable when show_feedback_to_mentor is actually set
+    # or this feature is actually ready?
+    # @feedbacks = current_user.solution_mentorships.where(show_feedback_to_mentor: true).order('updated_at desc').limit(5).pluck(:feedback)
+    @feedbacks = []
 
     if current_user.num_rated_mentored_solutions > Exercism::MENTOR_RATING_THRESHOLD
       @rating_threshold_reached = true


yyyc514:
Before:

```
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+-------------+
| id | select_type | table                | partitions | type | possible_keys       | key                 | key_len | ref   | rows | filtered | Extra       |
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | solution_mentorships | NULL       | ref  | fk_rails_578676d431 | fk_rails_578676d431 | 8       | const |    1 |    100.0 | Using where |
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+-------------+
```

After:

```
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+----------------------------------------------------+
| id | select_type | table                | partitions | type | possible_keys       | key                 | key_len | ref   | rows | filtered | Extra                                              |
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+----------------------------------------------------+
|  1 | SIMPLE      | solution_mentorships | NULL       | ref  | fk_rails_578676d431 | fk_rails_578676d431 | 8       | const |    1 |    100.0 | Using index condition; Using where; Using filesort |
+----+-------------+----------------------+------------+------+---------------------+---------------------+---------+-------+------+----------+----------------------------------------------------+
```

Of course I have no data, but I think the Extra is still informative.

yyyc514:
We can add an index also if needed.  Hard to say since I have no data locally.  The need to do sort here is going to require retrieving more data than previously so it can sort it before handing back the paged query.

yyyc514:
> Re your PR, the column you want is solutions.completed_at btw.

Isn't `solutions. updated_at` awfully close?  Esp if this is just going to change again in the future.  One index that covers BOTH these queries (from index) on the single critical table is going to be better than trying to sort solutions by a column you don't have an index for.

You're trading one fast (any tiny) query that can use the proposed index for one [HUGE one] that from where I'm sitting can't seem to use ANY index well.  (because you want to SCOPE by solution_mentorships, yet SORT by solutions).  It will use the user_id FK index against solution_memberships (cause it has no choice) but then still has to load JOIN with ALL the solutions so that it can then sort them in RAM by `completed_at`.

Does that make sense? Perhaps I need to upgrade my version of MySQL.  What version are we using in production?

iHiD:
On the user with the most solution_mentorships with feedback (you) it takes 10ms to run:
```ruby
u.solution_mentorships.with_feedback.joins(:solution).order('solutions.completed_at desc').limit(5).offset(300)
```

So I think we're ok without the index.

> Isn't solution_mentorships.updated_at awfully close? You're using it YOURSELF for this purpose on L74:

Yeah - but this is why people complain that they're not in a logical order, because if someone comes back later and adds a comment or iteration, it will then bump the feedback to the wrong place in the queue.

yyyc514:
> On the user with the most solution_mentorships with feedback (you) it takes 10ms to run:

Yeah, and I def want it to be fast for me. :-)  10ms is impressive.  I bet you'd be more impressed if you saw what it could do indexed. :-)  But I'll take your word it's "fast enough" for now then and go that direction with it.

> because if someone comes back later and adds a comment or iteration, it will then bump the feedback to the wrong place in the queue.

This is helpful, this is where you win with domain knowledge.  I wasn't quickly thinking of a case where the updated_at would change AFTER they left feedback... but adding more iterations (Which of course happens) makes sense... and yeah it would muck with ordering.

I'll try to swing back to this tomorrow.

Can I also turn on the "Celebration" feature that's currently disabled because of the `show_feedback_to_mentor`  field that is currently not used?

iHiD:
> I'll try to swing back to this tomorrow.

Thanks :)

> Can I also turn on the "Celebration" feature that's currently disabled because of the show_feedback_to_mentor field that is currently not used?

Yes, if you include a screenshot, as I can't remember what it looks like.

yyyc514:
No idea I've only seen the code, LOL.

yyyc514:
I didn't pull up the UI so I still have no idea what Celebration stream looks like.  I did disable the query in a separate commit here since we don't need it if we're never going to show it.  I can pull that back out if necessary but seems bad form to do extra work for no reason.

iHiD:
Thanks :) I'll deploy in a few mins.

PR #733, opened 2019-09-25
Use flipper to toggle mentor commentary

kntsoriano:
I've created a feature named `mentor_commentary` in `exercism.io/flipper`.

In that page, under actors, simply put in the user's ID. For clarity, `flipper_id` == `user_id`.


iHiD:
@kntsoriano When I try and add an actor id, I get the following:

Screenshot 2019-09-25 at 15 03 05

iHiD:
@kntsoriano I think the idiomatic way to do this is [this](https://github.com/exercism/website/blob/master/app/views/mentor/solutions/analysis/_javascript_resistor_color_duo_experiment.html.haml#L1) btw. Do you agree?

kntsoriano:
Yes. Looks a lot better!

PR #734, opened 2019-09-25
Remove specific styling for <pre> tags nested in <li> or <p>

kntsoriano:
Closes https://github.com/exercism/bugs/issues/29.

I think we've deliberately included a specific style for nested `<pre>` tags, but upon inspection it doesn't make that much of a difference. When deleting this specific style, nested `<pre>` tags in the dark mode theme work well.

### Without specific style
Screen Shot 2019-09-25 at 9 30 01 AM

### With style
Screen Shot 2019-09-25 at 9 30 13 AM

The only difference is a small padding to the left and we can add that if we wanted to.

### DIFF
--- a/app/assets/stylesheets/code-editor.scss
+++ b/app/assets/stylesheets/code-editor.scss
@@ -69,7 +69,7 @@ body .md-editor {
     }
 
     p,li {
-        > code {
+        &> code {
             border:1px solid #fafafa;
             background:#f5f5f5;
             @include code-font;
--- a/app/assets/stylesheets/code-editor.scss
+++ b/app/assets/stylesheets/code-editor.scss
@@ -69,7 +69,7 @@ body .md-editor {
     }
 
     p,li {
-        code {
+        > code {
             border:1px solid #fafafa;
             background:#f5f5f5;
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Perhaps you want to add the selectors with extra specificity: 

```css
body.prism-dark .editable-text .editable-text-html pre[class*="language-"], 
body.prism-dark .new-editable-text .preview-area pre[class*="language-"],
body.prism-dark .editable-text .editable-text-html li code,
body.prims-dark .new-editable-text .preview-area li code {

}
```

Would be a 2-line change.

kntsoriano:
Thanks, @SleeplessByte. I agree that the best solution in this case is to apply the style to inline code blocks only by using `>`. I think it was intended to behave like that in the first place, we just forgot to make the selector more specific.

app/assets/stylesheets/code-editor.scss
```diff
@@ -69,7 +69,7 @@ body .md-editor {
     }
 
     p,li {
-        code {
+        > code {
```
iHiD:
```suggestion
        &> code {
```

PR #735, opened 2019-09-25
Publish entry

kntsoriano:
https://github.com/exercism/wip/issues/35.

## Questions
1. Who is allowed to publish the entry? Should it only be the one who created the entry? With editing, in the spec, it's specified that users can only edit their own articles, should it be the same for publishing?
2. In the spec, there isn't anything about unpublishing an entry. I think we should include it as well. But it might get complicated quickly when we've already tweeted the links, and we might need to delete tweets as well.

### DIFF
--- a/app/controllers/changelog_admin/entries_controller.rb
+++ b/app/controllers/changelog_admin/entries_controller.rb
@@ -24,6 +24,14 @@ module ChangelogAdmin
       @entry = ChangelogEntry.find(params[:id])
     end
 
+    def publish
+      @entry = ChangelogEntry.find(params[:id])
+
+      @entry.publish!
+
+      redirect_to changelog_admin_entry_path(@entry)
+    end
+
     private
 
     def form_params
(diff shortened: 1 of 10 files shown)


iHiD:
> Who is allowed to publish the entry? Should it only be the one who created the entry? With editing, in the spec, it's specified that users can only edit their own articles, should it be the same for publishing?

In the spec I wrote "Admin users can see an extra button for "publish" on the show page, which, when clicked, sets the published_at.". I think that still make sense. So just `User#admin? == true` can see the publish button and publish.

Additionally, admins should also be able to edit things, as well as the creator.

> 2. In the spec, there isn't anything about unpublishing an entry. I think we should include it as well. But it might get complicated quickly when we've already tweeted the links, and we might need to delete tweets as well.

I think for now, if we want to delete something, we can just update the text. As you say, it will get tricky fast with tweets/emails etc, so best to just edit the text to say "Sorry - we've deleted this" or something, which we can do manually. When we've a bit more of an idea of how everything is working, we can add this.

kntsoriano:
I'll create a separate PR as this PR doesn't include anything about permissions yet.

PR #737, opened 2019-09-26
Edit entry

kntsoriano:

### DIFF
--- a/app/controllers/changelog_admin/entries_controller.rb
+++ b/app/controllers/changelog_admin/entries_controller.rb
@@ -29,6 +29,13 @@ module ChangelogAdmin
     def publish
       @entry = ChangelogEntry.find(params[:id])
 
+      unless AllowedToPublishEntryPolicy.allowed?(
+        user: current_user,
+        entry: @entry
+      )
+        return unauthorized!
+      end
+
       @entry.publish!
 
       redirect_to changelog_admin_entry_path(@entry)
--- a/app/controllers/changelog_admin/entries_controller.rb
+++ b/app/controllers/changelog_admin/entries_controller.rb
@@ -42,9 +42,14 @@ module ChangelogAdmin
       @form = ChangelogEntryForm.from_entry(@entry)
 
       @form.assign_attributes(form_params)
-      @form.save
 
-      redirect_to changelog_admin_entry_path(@entry)
+      if @form.valid?
+        @form.save
+
+        redirect_to changelog_admin_entry_path(@entry)
+      else
+        render :edit
+      end
     end
 
     private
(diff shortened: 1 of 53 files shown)


app/forms/changelog_entry_form.rb
```diff
@@ -5,6 +5,12 @@ def initialize(msg = "User isn't allowed to save an entry")
     end
   end
 
+  class CantChangeCreatedByError < StandardError
```
iHiD: See comment above about permit / strong params here.

app/forms/changelog_entry_form.rb
```diff
@@ -30,11 +36,16 @@ def self.from_entry(entry)
...
     :referenceable_gid,
     :info_url,
-    :created_by
+    :created_by,
   )
 
+  attr_reader :created_by
+
   def save
     raise UnauthorizedUserError unless created_by.may_edit_changelog?
```
iHiD: This is problematic, because if a user has their changelog permissions removed, then no admin can ever change the changelog. I would think that `created_at` here should be the actual user changing it, which should probably be passed in as a param if this is where the security is happening now.

app/controllers/changelog_admin/entries_controller.rb
```diff
@@ -51,7 +51,12 @@ def update
     def check_if_entry_is_editable!
       @entry = ChangelogEntry.find(params[:id])
 
-      return unauthorized! if @entry.created_by != current_user
+      unless AllowedToEditEntryPolicy.allowed?(
+        user: current_user,
+        entry: @entry
+      )
+        return unauthorized!
```
iHiD: I'd postfix this, so 
```ruby
return unauthorized! unless AllowedToEditEntryPolicy.allowed?(
        user: current_user,
        entry: @entry
      )
```

app/controllers/changelog_admin/entries_controller.rb
```diff
@@ -32,13 +34,41 @@ def publish
...
     end
 
+    def edit
+      @form = ChangelogEntryForm.from_entry(@entry)
+    end
+
+    def update
+      @form = ChangelogEntryForm.from_entry(@entry)
+
+      @form.assign_attributes(form_params)
```
iHiD: Is there a reason not to use permit from Strong Paramaters here to stop a change to current_user, rather than having all the explict checks for it in the form model? Especially as we're guarding against malicious intent here. I feel like we risk rewriting stuff that we already get for free in Rails. Although I might just be misunderstanding.

kntsoriano: Thinking about it, I agree that we should just use strong parameters. I imagine that this model won't be used outside the request model and it would be less complicated if we use strong parameters. Thanks :)

iHiD: Awesome :)

PR #740, opened 2019-09-26
Save reference key

kntsoriano:


app/forms/changelog_entry_form.rb
```diff
@@ -69,4 +70,10 @@ def entry
   def details_html
     ParseMarkdown.(details_markdown)
   end
+
+  def referenceable_key
+    return if referenceable.blank?
+
+    "#{referenceable.class}_#{referenceable.id}"
```
iHiD: Could you underscore this pls? `referenceable.class.name.underscore`

PR #741, opened 2019-09-27
View changelog

kntsoriano:

## Questions
1. Should we include the time the entry was published in here as well?
2. In the upcoming show page, should the URLs be slugs instead of ids? (e.g. `exercism.io/changelog/new-exercise-added`)


app/models/changelog_admin/null_referenceable.rb
```diff
@@ -0,0 +1,15 @@
+module ChangelogAdmin
+  class NullReferenceable
+    def icon
+      "https://https://via.placeholder.com/150"
```
kntsoriano: This is the default icon when nothing is referenced in the changelog entry.

app/models/changelog_admin/referenceable_exercise.rb
```diff
@@ -1,13 +1,19 @@
...
 
-    private
+    def title
+      "#{exercise.track_title} - #{exercise.title}"
+    end
 
-    def exercise
-      __getobj__
+    def icon
+      exercise.dark_icon_url
```
kntsoriano: I'm unsure which icon to use.

app/models/changelog_admin/referenceable_track.rb
```diff
@@ -0,0 +1,19 @@
...
+    def initialize(track)
+      @track = track
+    end
+
+    def title
+      track.title
+    end
+
+    def icon
+      track.bordered_green_icon_url
```
kntsoriano: Same here. Unsure of which icon to use.

app/controllers/changelogs_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ChangelogsController < ApplicationController
```
iHiD: Could we call this `ChangelogEntriesController` pls?

app/models/changelog_admin/null_referenceable.rb
iHiD:
```suggestion
      "https://via.placeholder.com/150"
```

iHiD: OK - cool. I'll work this out with Nic when we do styling.

config/routes.rb
```diff
@@ -132,6 +132,8 @@
     end
   end
 
+  resource :changelog, only: [:show]
```
iHiD: I think this should be resources `:changelog_entries, only: [:index], path "changelog"`

kntsoriano: Sure. Will do.

kntsoriano:
@iHiD Any thoughts on the questions above?

iHiD:
@kntsoriano Missed those. Sorry.

1. Should we include the time the entry was published in here as well?

Yeah. That'd be nice. Using `time_ago_in_words` pls.

2. In the upcoming show page, should the URLs be slugs instead of ids? (e.g. exercism.io/changelog/new-exercise-added)

Good question. I like slugs, but I can see loads of slug clashes. Could you use slugs but posfixed with the id pls? So `exercism.io/changelog/new-exercise-added-#{id}`. That should be straight forward in FriendlyId I think.

PR #747, opened 2019-10-01
Update README to reflect windows release

sjwarner-bp:
Windows 1809 has been released now.

It's worth updating the README here to reflect that. Is it also worth removing the tracking issue?


iHiD:
Thanks. I'm not sure which the tracking issue is?

PR #750, opened 2019-10-02
Appending parenthesized count of solutions one is mentoring to Mentor Dashboard item in the .misc_menu

n8chz:
I thought it might be convenient for mentors to have a preview of how many solutions they have in-process before actually going to mentor/dashboard/your_solutions

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,3 +87,5 @@ end
 group :development, :production do
   gem 'capistrano-checks'
 end
+
+gem 'pry-rails'
--- a/Gemfile
+++ b/Gemfile
@@ -87,5 +87,3 @@ end
 group :development, :production do
   gem 'capistrano-checks'
 end
-
-gem 'pry-rails'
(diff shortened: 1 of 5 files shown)


n8chz:
I tried adding a link to the "New Solutions" page, visible in cases where "Solutions You're Mentoring" is empty, but without a count of solutions. Logic is, skip SYM page if cleared, one less click, but I don't know how many more milliseconds wait.

sshine:
> Is there a way to use pry by default in a working copy w/o putting pry gem in commits?

Don't commit them?

You could put the part that adds pry into a .diff file and apply it with `patch -p1 < pry.diff` whenever you need to; this means you can clear pry-related changes with `git checkout`. Or you could put your pry-related changes in a feature branch and make an alias to `git cherry-pick` on the HEAD of that particular feature branch.

Then the only obstacle is to not add those lines when committing real stuff; beyond not using `git add .`, it depends on your workflow and git interface (CLI or IDE plugin). I personally use `git add -p` to avoid submitting things that were not meant for a PR, but there are several ways to achieve this.

PR #758, opened 2019-10-04
Add published at to changelog view

kntsoriano:


test/system/guest_views_changelog_test.rb
```diff
@@ -2,6 +2,7 @@
 
 class GuestViewsChangelogTest < ApplicationSystemTestCase
   test "guest views changelog" do
+    travel_to(Time.utc(2016, 12, 26))
```
iHiD: Can we wrap this in a block, rather than having the `travel_back` at the bottom, so we don't get time mishaps if there's an exception?

kntsoriano: No problem :) Agree with using the block syntax as it's safer.

PR #759, opened 2019-10-04
Allow admin to choose an empty referenceable

kntsoriano:


app/views/changelog_admin/entries/_form.html.haml
```diff
@@ -21,7 +21,8 @@
     :all,
     :name,
     :to_global_id,
-    :title
+    :title,
+    { prompt: "None" }
```
iHiD:
```suggestion
    { prompt: "General" }
```

PR #760, opened 2019-10-04
Unpublish entry

kntsoriano:

This assumes the ff:
1. Dead links in tweets are fine for now. We can delete tweets later by storing the tweet ID in the database, which we don't have yet.
2. Only admins are allowed to unpublish entries.
3. A new tweet will be sent out if admin chooses to republish the entry.


app/view_models/changelog_admin/changelog_entry_action.rb
```diff
@@ -0,0 +1,44 @@
+module ChangelogAdmin
+  class ChangelogEntryAction
+    def self.all
+      [
+        ChangelogEntryAction.new(name: :publish, confirm: true, method: :post),
+        ChangelogEntryAction.new(name: :unpublish, confirm: true, method: :post),
```
iHiD: Can we comment this out for now pls?

kntsoriano: Yeah. Makes sense. I'll comment out and merge.

PR #762, opened 2019-10-04
Delete entry

kntsoriano:

Building on https://github.com/exercism/website/pull/760, this allows admins to delete entries. 

Entries are allowed to be deleted if: (1) a site admin does it or (2) a changelog admin does it provided that the entry is still unpublished.


kntsoriano:
@iHiD As suggested in #760, I'll comment out this feature as well.

PR #763, opened 2019-10-07
Add fallback profile image in team pages

kntsoriano:
Closes https://github.com/exercism/wip/issues/42.


app/helpers/profile_image_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ProfileImageHelper
+  def profile_image(person)
```
iHiD: Could we call this `ContributorImage` (and `ContributorImageHelper`) please. `Profile` refers to user's profiles and I think we have other helper methods defined for that already, so the naming could get complex.

kntsoriano: Thanks :) Didn't really know what to call this.

PR #766, opened 2019-10-08
Add basic styling to changelog admin

iHiD:
I've added some basic styling. We'll style it more properly when we have some more time but this gets it to a point where it's not blocked :)

Screenshot 2019-10-08 at 15 48 56

### DIFF
--- a/app/assets/stylesheets/changelog_admin/edit-entry.scss
+++ b/app/assets/stylesheets/changelog_admin/edit-entry.scss
@@ -1,0 +1,16 @@
+@import 'variables';
+@import 'mixins';
+
+#changelog-admin-edit-entry {
+    table {
+        margin-bottom:10px;
+        th, td {
+            border:1px solid #ddd;
+            padding:10px;
+        }
+    }
+    a {
+        color:$color-2;
+    }
+}
+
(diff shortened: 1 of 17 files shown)


iHiD:
@kntsoriano If you get chance, could you add a commit fixing the tests in this PR? If not, I'll do it when I get chance on Thursday my time. Thanks.

iHiD:
> Would it be worth it to attach classes to the elements instead of selecting by tag name in order to make unexpected change happen?

So this makes sense, but at the same time, if we're going to use semantic things like `dt` and `dd`, it feels like adding styling for those elements makes sense. Same with `table`, `h1` and `p` tags. You're probably right about not using `.pure-button`, but the whole codebase has that right now, so we should decide exactly how we're going to change it to be consistent if we want to.

Maybe we have a day in Tokyo where we decide on some rules for all this that you can make sure I stick to moving forward? :)

kntsoriano:
Yes. I agree that semantic styling makes sense, but from my experience in our codebase it was really hard for me to add new styles because the CSS was generally placed. I might be using the wrong tags, though.

A day to catch up on this would be good :)

PR #767, opened 2019-10-08
Optimise images for about pages

iHiD:
@nicolechalmers This reduces the image sizes on the about pages by 90%. We should try to remember to use jpgs for photos as they're much more efficient. Please check that you don't feel the image quality has degraded on any of those pages :)


nicolechalmers:
@iHiD They all still look great. Thanks for doing that. I'll try to remember for next time.

PR #770, opened 2019-10-09
Add markdown preview

kntsoriano:
Closes https://github.com/exercism/wip/issues/44.

I decided to repurpose our markdown preview in discussion posts as it seems to be easily pluggable.

The only problem I've noticed are that header tags aren't rendered properly as the CSS was maybe written to accommodate the comments use case. If this is okay with us, we can merge this, otherwise, I'll be injecting a different CSS to render the header tags correctly.

## Markdown
Screen Shot 2019-10-09 at 1 52 59 PM
## Rendered
Screen Shot 2019-10-09 at 1 53 03 PM


iHiD:
Thanks dude.

PR #781, opened 2019-10-18
Add test results via websockets POC

kntsoriano:
This PR shows a proof of concept of how test results are delivered via websockets.

## Discussion
1. In order to support the listening daemon to send messages to our ActionCable channels, we'd need to use the redis adapter provided by ActionCable. The default adapter is in-process only and not recommended for production use. Furthermore, Rails supports another adapter for PostgreSQL, but I think we should go with redis given how are infrastructure is right now.
2. Rails recommends that we render all the HTML server side vs. client side. In DHH's screencast on ActionCable, he renders the HTML in a background job via the external renderer available. I think we should do the same as we are used to having our views rendered server side and client side rendering might be harder to handle for us. In this proof of concept, I haven't done this yet and just opted to copy paste the existing code into the proper ActionCable handler.
3. Overall, I feel positive about this direction. ActionCable seems to be a core part of Rails and support for it is good. It still feels like writing Rails with just a few concepts to brush up on.


iHiD:
@kntsoriano Thank you!
@ccare You have questions (concerns? / thinks to check?) about how websockets works in Rails that you've previously asked me. Could you ask them here so @kntsoriano can investigate/confirm/education us pls :)

app/services/testing_services/process_tested_submission.rb
```diff
@@ -18,6 +18,22 @@ def call
...
+      status = if submission.tested?
+                 submission.test_results.last.results_status
+               else
+                 :failed_dependency
+               end
+
+      message = submission.test_results.last.message if status == :error
+      failure = submission.test_results.last.tests.select{|t|t['status'] == 'fail'}.first if status == :fail
+
+      SubmissionChannel.broadcast_to(
```
ccare: does `broadcast_to` mean that all connected browsers will receive the message?

ccare: I assume this works across process  - assuming that service is invoked via a rake-based propono listener, then it must be. Does that mean that the plumbing is all handled by redis? I assume that means we need a single redis for both webapps and processor to share in order for delivery to work to any connected consumer.

ccare: If we have a single redis for all of this, do we want to deploy it standalone (so that it's an isolated single point of failure) from everything else?

ccare: It would be useful to understand what would happen if redis were to go rogue and we couldn't deliver messages. Would the web UI continue to work (albeit with the browser needing the occassional refresh). I guess I'm asking whether websockets are an enhancement or a critical component, because that informs decisions about resilience of the redis broker

ccare: Do we have any idea how transient the subscription-data in redis is? I imagine it's pretty transient but it would be good to confirm.

kntsoriano: I would probably say that websockets are an enhancement in order to deliver the data without a page refresh. I think we should still save everything to the database so that when the user refreshes they would get the same page as with when websockets delivers the data.

I'm not sure how long the timeout is before Rails cleans up subscription information. The only thing I know is that each subscription is an open pub sub channel between redis and the server and would be maintained the moment we'd need more messages.

Again, I think these questions are also a matter of how we would want the UI to be. If we aren't saving the build runs as what I've said above, we might have lots of long running connections between each user solving a problem versus a lot more short-lived connection for each user when modelling it as a separate channel for each "build". It would good to discuss what the pros and cons are for each and what user experience we are aiming for.

iHiD:
> @kntsoriano: Furthermore, Rails supports another adapter for PostgreSQL, but I think we should go with redis given how are infrastructure is right now.
> @ccare: If we have a single redis for all of this, do we want to deploy it standalone (so that it's an isolated single point of failure) from everything else?

I'd like to go with a hosted solution to this. Every time we've put Redis in production we've hit some sort of problems managing it, and I think it's an issue that you both can do without having to worry about. So I vote for either a hosted Redis (if there is such a thing - [Elasticache Redis](https://aws.amazon.com/redis/) maybe?) or RDS Postgres.

iHiD:
> I haven't done this yet and just opted to copy paste the existing code into the proper ActionCable handler.

Cool. So I think for the next commit on this PR, it would be good to see how that is handled.

iHiD:
Having read the docs, I am unsure why we don't have/need a connection. e.g.:

````# app/channels/application_cable/connection.rb
module ApplicationCable
  class Connection < ActionCable::Connection::Base
    identified_by :current_user
 
    def connect
      self.current_user = find_verified_user
    end
 
    private
      def find_verified_user
        if verified_user = User.find_by(id: cookies.encrypted[:user_id])
          verified_user
        else
          reject_unauthorized_connection
        end
      end
  end
end```

Thoughts?

ccare:
This was where I was going with my question of how transient the data is.
If we don't need to maintain our own backups of whats in Redis, then using
elasticache is a good call.

PR #784, opened 2019-10-22
Fix migration modal typo

ngsikai:
Fix copy for migration modal as suggested here https://github.com/exercism/exercism/issues/5040 😄


iHiD:
Sweet. Thanks! :)

PR #790, opened 2019-10-22
Fix code samples in dark theme mode

iHiD:
This builds on #567 to improve the way we render dark theme throughout the site. it DRYs up the code further and ensures that the dark-theme box fits neatly into the graphic.

Closes exercism/exercism#4690

Screenshot 2019-10-22 at 17 59 22
Screenshot 2019-10-22 at 17 59 40

### DIFF
--- a/app/assets/stylesheets/pages/solution-page.scss
+++ b/app/assets/stylesheets/pages/solution-page.scss
@@ -61,7 +61,6 @@
 
     .test-suite,
     .solution.multi-file {
-        background:#f9f9f9;
         h3 {
             padding:10px 50px 10px 27px;
             font-weight:$bold;
@@ -79,9 +78,6 @@
             margin-right: 10px;
         }
     }
-    .test-suite pre {
-        background:#f9f9f9;
-    }
     .solution pre {
         background:#fff;
     }
@@ -90,6 +86,7 @@
         pre {
             padding-top:0;
             padding-bottom:0;
+            border-radius: 0;
         }
     }
     .lhs-content {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/prism-modifications.scss
```diff
@@ -35,7 +35,3 @@ pre.line-numbers {
         }
     }
 }
-
-pre[class*="language-"] {
-  border-radius: 0 !important;
```
iHiD: This shouldn't be global. I've moved it to the relevant place in the solution css file.

app/assets/stylesheets/pages/track-page.scss
```diff
@@ -157,24 +158,23 @@ body.controller-my-tracks.action-show {
             background-image:image-url("code-sample-frame.png");
             background-size:contain;
             background-repeat:no-repeat;
-            padding:30px 10px;
+            padding:28px 6px;
             height:200px;
             width: 258px;
 
             pre {
-                background:#fafafa;
```
iHiD: This file shouldn't need to set background colors. It should be done in one central place in the solution file.

app/assets/stylesheets/pages/track-page.scss
```diff
@@ -157,24 +158,23 @@ body.controller-my-tracks.action-show {
...
                 @include font-size(10, 13);
-                padding:2px 5px;
-                height:144px;
-                code {
-                    background:#f8f8f8;
-                }
+                padding:10px 10px;
+                height:151px;
+                margin:0;
+                border-radius: 0 0 9px 9px;
```
iHiD: As the dark background now cuts against the graphic, I've tidied the padding and border radiuses to make it fit neatly.

app/assets/stylesheets/pages/solution-page.scss
```diff
@@ -362,20 +359,36 @@ body:not(.prism-dark) #teams-my-solution-page {
     }
 }
 
-body.prism-dark {
+body:not(.prism-dark) {
```
kntsoriano: I'm not sure if this is the right place for this. I think we should extract a separate component for the prism box soon and handle themes in there. I'm okay with this for now as it is how we have things right now.

iHiD: > I'm not sure if this is the right place for this. 

Agree. But I think this then becomes the only file we use it in (We already use it lots in this file already)

>  think we should extract a separate component for the prism box soon and handle themes in there

I couldn't agree more.

PR #793, opened 2019-10-23
Remove Perl syntax highlighting for Perl 6

mienaikage:


iHiD:
@mienaikage Could you explain what this is doing and why please? Specifically what problem do you have that it is solving? 

---

cc @sshine as you did the opposite of this commit last: https://github.com/exercism/website/commit/5bea07694bf260a4764ea1aa14c33734d51b97e2#diff-d709783f9d1b42e7a3ce9a711152c562 

---

cc @kntsoriano as the build is failing with the following (I presume unrelated) error:
```
Error:
FilterSolutionsTest#test_filters_solutions_by_exercise:
Capybara::ElementNotFound: Unable to find css "div.option" within #<Capybara::Node::Element tag="div" path="/HTML/BODY[1]/DIV[1]/DIV[2]/DIV[1]/DIV[1]/FORM[1]/DIV[3]/DIV[2]">
    test/support/selectize_helpers.rb:12:in `block in select_option'
    test/support/selectize_helpers.rb:10:in `select_option'
    test/system/mentor/filter_solutions_test.rb:117:in `block in <class:FilterSolutionsTest>'
rails test test/system/mentor/filter_solutions_test.rb:95
```

mienaikage:
I'd originally come to this file to remove the Perl syntax highlighting for Perl 6. While they share some similarities, they are ultimately different languages, and the highlighting doesn't suit Perl 6 too well e.g. kebab case is very common in Perl 6, but isn't highlighted properly. There is currently an issue open in Prism to support syntax highlighting for Perl 6 which has not yet been implemented.

The changes for Perl 5 came from poor assumptions from when I'd checked the file in the past. I'll cut those out.

sshine:
@iHiD: This seems PR safe since @mienaikage has implicitly assessed that the Perl 6 (aka Raku) track should not have syntax highlighting due to the current lack of support in Prism.

iHiD:
Thanks both! :)

PR #804, opened 2019-11-26
Show submission status via websockets

kntsoriano:


config/cable.yml
```diff
@@ -1,10 +1,11 @@
...
+  adapter: redis
+  url: redis://localhost:6379/1
 
 test:
   adapter: test
 
 production:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= Rails.application.secrets.action_cable_redis_url || "redis://localhost:6379/1" %>
```
iHiD: Can we get rid of this `||` and just use the secrets. I don't want to accidentally use a different localhost.

PR #807, opened 2019-11-28
Make submissions.solution polymorphic

iHiD:


kntsoriano:
@iHiD What do we need this for?

iHiD:
Because TeamSolution and Solution (and ExperimentSolution coming soon) all share the same ids, so we need this to be polymorphic so we know what type of Solution this is pointing to. This is the same thing we do for Iterations for example.

PR #810, opened 2019-11-29
omit current page from "More $LANGUAGE Resources"

iHiD:
By @davearonson:

This fixes issue [#993 from the website-copy repo](https://github.com/exercism/website-copy/issues/993), in which jj-plane complained that the Installing Python page links back to itself.  That is not limited to that link on that page, but _all_ the language-specific links in that list on such pages in _all_ language tracks.

This solution _omits_ the current page's link out of the list.  Another alternative would be to make it not a link (such as by using `link_to_unless_current`). Even then, though, the list header stating "More $LANGUAGE Resources" would still be misleading, as the item listed (whether linked or not) would not be "More".


sshine:
Should this be merged?

kntsoriano:
@iHiD Can we merge this?

PR #812, opened 2019-11-29
Add flow for joining research experiments

iHiD:
@kntsoriano This handles the first steps beyond what you're doing with the joining research, in letting someone join an actual experiment.

You might want to rebase your stuff on top of this, or cherry pick this commit onto your branch. 

Either way, probably better to do that sooner rather than later :)

### DIFF
--- a/app/controllers/research/user_experiments_controller.rb
+++ b/app/controllers/research/user_experiments_controller.rb
@@ -1,0 +1,20 @@
+module Research
+  class UserExperimentsController < Research::BaseController
+    def create
+      experiment = Experiment.find(params[:experiment_id])
+      begin
+        UserExperiment.create!(
+          user: current_user,
+          experiment: experiment
+        )
+      rescue ActiveRecord::RecordNotUnique
+      end
+      redirect_to research_user_experiment_path(experiment)
+    end
+
+    def show
+      @experiment = Experiment.find(params[:id])
+      @user_experiment = UserExperiment.find_by(user: current_user, experiment: @experiment)
+    end
+  end
+end
--- a/app/models/concerns/solution_base.rb
+++ b/app/models/concerns/solution_base.rb
@@ -61,6 +61,18 @@ module SolutionBase
     true
   end
 
+  def team_solution?
+    false
+  end
+
+  def experiment_solution?
+    false
+  end
+
+  def use_auto_analysis?
+    false
+  end
+
   #class_methods do
   #  ...
   #end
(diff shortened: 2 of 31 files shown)


app/controllers/research/user_experiments_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Research
+  class UserExperimentsController < Research::BaseController
+    def create
+      experiment = Experiment.find(params[:experiment_id])
+      begin
```
kntsoriano: Why do we have this wrapped?

test/models/research/experiment_solution_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
```
kntsoriano: Can we remove these empty tests?

test/system/research/join_experiment_flow_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  class JoinExperimentFlowTest < Research::TestCase
+    test "user joins an experiment from the experiments page" do
+      user = create(:user, :onboarded)
+      experiment = create(:research_experiment)
+
+      sign_in!(user)
+      visit research_experiments_path
+      click_on "Participate"
+
+      assert UserExperiment.where(user: user, experiment: experiment).exists?
```
kntsoriano: I think we should assert at the integration level, preferably checking text exists on this page, etc.

app/models/concerns/solution_base.rb
```diff
@@ -61,6 +61,18 @@ def latest_exercise_version?
     true
   end
 
+  def team_solution?
```
kntsoriano: I'm a bit scared of this and what this would imply in the future. I'm okay with this right now, but asking the solution what kind of solution it is doesn't seem like good design as it's the same as asking what class the object is.

Let's see where we get to and refactor later.

app/controllers/research/user_experiments_controller.rb
iHiD: To stop race conditions when people double click (See the test for it).

app/models/concerns/solution_base.rb
iHiD: Yeah, agree. That method is already used throughout the codebase though. This is just a DRYing up of the existing code to move it into one place.

test/system/research/join_experiment_flow_test.rb
iHiD: I sort of agree. But we don't have any text yet that won't change, and I didn't really want to tightly-couple the test to the wrong text. Thoughts?

test/models/research/experiment_solution_test.rb
iHiD: Sure. Want to do it when you cherry-pick?

PR #817, opened 2019-12-02
Download submissions

iHiD:
Closes https://github.com/exercism/internal-wip/issues/17

This does two things:
1) It stores the filenames as uuids for security, so we don't allow arbitary paths in our s3.
2) It downloads the files on demand in a multithreaded way.

### DIFF
--- a/app/services/submission_services/download_files.rb
+++ b/app/services/submission_services/download_files.rb
@@ -1,0 +1,45 @@
+module SubmissionServices
+  class DownloadFiles
+    include Mandate
+    def initialize(submission_uuid, filenames)
+      @filenames = filenames
+      @files = Concurrent::Map.new
+      @path = "#{Rails.env}/storage/#{submission_uuid}"
+    end
+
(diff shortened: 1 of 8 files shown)


app/services/submission_services/download_files.rb
```diff
@@ -0,0 +1,45 @@
...
+    def call
+      threads = []
+      filenames.each do |original_filename, stored_filename|
+        threads << Thread.new { download_file(original_filename, stored_filename) }
+      end
+      threads.each(&:join)
+
+      # Convert to a normal hash for the return.
+      # There's probably a nicer way to do this but
+      # RubyDoc is down for me so :shrug:
```
kotp: The official documentation can be found https://docs.ruby-lang.org/en/ instead of RubyDoc.  The documentation can also be generated from the source locally.

iHiD: @kotp Thank you.

PR #819, opened 2019-12-03
Use language_track, not research_track, when submitting for test-running

iHiD:
When we pass the track slug to the test runners we should use the language track slug (e.g. `csharp`) not the "fake" research tracks where the research exercises actually live.

Note: The extraction of `Research` into modules was done because ruby defined `Research` as a class if it came across it in the context of `class Research::Foo` before `module Research`

### DIFF
--- a/app/controllers/research/experiment_solutions_controller.rb
+++ b/app/controllers/research/experiment_solutions_controller.rb
@@ -3,12 +3,12 @@ module Research
     def create
       experiment = Experiment.find(params[:experiment_id])
 
-      # TODO - Remove this placeholder
-      exercise = Exercise.find_by(slug: "two-fer", track: Track.find_by_slug("ruby"))
+      exercise_slug = "#{params[:language]}-#{params[:part]}-#{%w{a b}.sample}"
+      exercise = Exercise.find_by_slug!(exercise_slug)
 
-      # TODO - Check this is on the experiment track etc
-      #exercise_slug = "#{params[:language]}_#{%w{a b}.sample}_#{part}"
-      #exercise = Exercise.find_by_slug!(exercise_slug)
+      # Guard to ensure that someone doesn't try and access
+      # a non-research solution through this method.
+      raise "Incorrect exercise" unless exercise.track.research_track?
 
       solution = Research::CreateSolution.(
         current_user,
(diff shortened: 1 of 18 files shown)


test/services/submission_services/run_tests_test.rb
```diff
@@ -12,5 +12,20 @@ class RunTestsTest < ActiveSupport::TestCase
...
     end
+
+    test "uses language_track" do
+      research_track = create :track, slug: "research_123"
+      ruby_track = create :track, slug: "ruby"
+      exercise = create :exercise, track: research_track, slug: "ruby-1-b"
+
+      submission = create :submission, solution: create(:research_experiment_solution, exercise: exercise)
+
+      RestClient.expects(:post).with('http://test-runner.example.com/submissions',
```
ErikSchierboom: Just to be safe: research solutions are put in a different mentoring queue from the regular one, right?

ErikSchierboom:
@iHiD The build failed.

test/services/submission_services/run_tests_test.rb
iHiD: It's a different type of solution (`Research::ExperimentSolution`), so they are treated differently with regards to mentoring queues.

For test-running, the only thing using test running right now is the experiments, so there's no overlap.

ErikSchierboom: Cool! The PR looks great

app/controllers/research/experiment_solutions_controller.rb
```diff
@@ -3,12 +3,12 @@ class ExperimentSolutionsController < Research::BaseController
...
-      exercise = Exercise.find_by(slug: "two-fer", track: Track.find_by_slug("ruby"))
+      exercise_slug = "#{params[:language]}-#{params[:part]}-#{%w{a b}.sample}"
+      exercise = Exercise.find_by_slug!(exercise_slug)
 
-      # TODO - Check this is on the experiment track etc
-      #exercise_slug = "#{params[:language]}_#{%w{a b}.sample}_#{part}"
-      #exercise = Exercise.find_by_slug!(exercise_slug)
+      # Guard to ensure that someone doesn't try and access
+      # a non-research solution through this method.
+      raise "Incorrect exercise" unless exercise.track.research_track?
```
ErikSchierboom: I like this guard!

PR #820, opened 2019-12-03
Split view

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/15.

## Description
- This already has ActionCable integration. The status is updated as the submission moves through the queue.
- I decided to render the status on the client side because we have the "Queueing" state that is only client side.
- Test results are rendered server side and the HTML is sent via ActionCable.

## Screenshots
Screen Shot 2019-12-03 at 5 05 25 PM
Screen Shot 2019-12-03 at 5 05 16 PM
Screen Shot 2019-12-03 at 5 05 09 PM
Screen Shot 2019-12-03 at 5 05 01 PM
Screen Shot 2019-12-03 at 5 04 53 PM


iHiD:
Awesome. I'm going to merge this so we can try it, and then I'll read it through properly later :)

PR #832, opened 2019-12-05
Integrate test-runner results with SPI

iHiD:


iHiD:
@kntsoriano I'll merge this when it goes green, but cc'ing you FYI.

PR #834, opened 2019-12-05
Submit solution to researchers

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/23.

### DIFF
--- a/app/models/research/experiment_solution.rb
+++ b/app/models/research/experiment_solution.rb
@@ -43,5 +43,13 @@ module Research
 
       files
     end
+
+    def submit!(time: Time.current)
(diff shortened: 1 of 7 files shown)


app/models/research/experiment_solution.rb
```diff
@@ -43,5 +43,13 @@ def latest_files
 
       files
     end
+
+    def submit!(time: Time.current)
```
kntsoriano: We might want to guard against submitting solutions that haven't passed the tests.

iHiD:
This is great, thanks. I'll add a commit renaming `finished_at` to `submitted_at`.

PR #839, opened 2019-12-05
Fix extra whitespaces

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/24.


app/views/research/experiment_solutions/_solution.html.haml
```diff
@@ -1,6 +1,5 @@
 - solution.latest_files.each do |filename, code|
-  .code-editor.js-code-editor{data: { language: solution.language_slug, filename: filename } }
-    = code
+  .code-editor.js-code-editor{data: { language: solution.language_slug, filename: filename } }= code
```
iHiD: @kntsoriano is it worth adding a test for this. It would be a really easy thing to regress. Maybe make a note that when everything is stable, this is something we write a test to check?

PR #851, opened 2019-12-09
Add shortcuts

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/30.


iHiD:
@kntsoriano Is `?` now supposed to open the keyboard shortcut model? If so, it doesn't work for me.

PR #859, opened 2019-12-10
Broadcast latest submission

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/35.

### DIFF
--- a/app/jobs/broadcast_submission_job.rb
+++ b/app/jobs/broadcast_submission_job.rb
@@ -1,22 +1,5 @@
 class BroadcastSubmissionJob < ApplicationJob
   def perform(submission)
-    test_run_html = ApplicationController.render(
-      partial: submission.test_run,
-      as: :test_run
-    )
-
-    ops_status_html = ApplicationController.render(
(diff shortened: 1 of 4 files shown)


app/jobs/broadcast_submission_job.rb
```diff
@@ -1,22 +1,5 @@
...
-
-    ResearchSolutionChannel.broadcast_to(
-      submission.solution,
-      {
-        opsStatus: submission.ops_status,
-        opsStatusHtml: ops_status_html,
-        testRunHtml: test_run_html,
-      }
-    )
+    BroadcastSolutionJob.perform_now(submission.solution)
```
iHiD: Can we stop this indirection and just have `BroadcastSolutionJob`?

iHiD:
What happens if someone submits two submissions and
- Broadcast queued (s1)
- Broadcast queued (s2)
- Broadcast starting (s1)
- Broadcast finished (s1)
- Broadcast starting (s2)

How will the code respond to the final "Broadcast starting"?

PR #860, opened 2019-12-10
Small refactor

iHiD:
@kntsoriano I found this code quite hard to read / reason about so I've tried to simplify it.

Best to read each commit in order - the messages should explain what they do and why. Let me know what you think :)


kntsoriano:
I've merged this to my PR.

PR #873, opened 2020-01-06
Fixes median wait times

iHiD:
@kntsoriano There was a bug in the median wait time calculations. I've refactored by removing a load of the SQL and just calculating the medians in Ruby. It simplifies the code and fixes the bug.

Annoyingly, I don't know what caused the bug and can't get a failing test for it, but it was very much wrong in production. 

I'm self-merging/deploying this as it has the side effect of marking the wrong tracks as oversubscribed.

---

Update: I've also added another commit to reset the wait time if there are no solutions that match.

### DIFF
--- a/app/services/track_services/update_median_wait_time.rb
+++ b/app/services/track_services/update_median_wait_time.rb
@@ -3,59 +3,11 @@ module TrackServices
     include Mandate
 
     def call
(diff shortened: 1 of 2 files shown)


kntsoriano:
@iHiD Thanks! I'll do a better job of making sure everything works next time.

PR #884, opened 2020-01-23
Append class to page based on submission status

kntsoriano:


app/javascript/src/experiment-solution.js
```diff
@@ -101,12 +101,40 @@ class ExperimentSolution {
...
+          break;
+        case 'cancelling':
+          this.updateClass('info');
+          break;
+        case 'cancelled':
+          this.updateClass(previousStatus);
+          this._cancelBuild();
+          break;
+        case 'tested':
+          this.updateClass('tested');
```
iHiD: Should this be `pass` or `fail`?

kntsoriano: If we want to have a distinction between pass or fail we can add it. Currently we just let the submission render itself whether it passes or fails.

PR #889, opened 2020-01-24
Switch page theme

kntsoriano:
https://github.com/exercism/internal-wip/issues/41.

This PR switches the page's theme based on what is selected. This changes the editor's theme, PrismJS theme, and the body class.

The also saves the selected theme to the database via AJAX and sets the saved theme whenever the page loads.

I needed to refactor the PrismJS stylesheets in order to load both themes. They will be toggled by having the `disabled` attribute.


iHiD:
This is really great. Thanks dude. I've added one commit to move the controller stuff into the preferences_controller where this functionality already exists.

I've also pushed another commit to try and get it green, but I don't understand why it's not, as the HAML that the linter is failing on is valid.

app/javascript/src/code-editor.js
```diff
@@ -12,11 +12,7 @@ class CodeEditor {
   _setup() {
     this.editor = new AceEditor(this.element, this.config, this.filename);
 
-    this.editor.onSetup = () => {
-      this.editor.setTheme('dark');
-
-      this.onSetup(this);
-    }
+    this.editor.onSetup = () => { this.onSetup(this); }
```
iHiD: Where is `this.onSetup(this)` defined? I don't understand this line :)

PR #900, opened 2020-01-30
Test results

kntsoriano:
Closes https://github.com/exercism/research_experiment_1/issues/60.

### DIFF
--- a/app/views/research/submission_test_run_results/_fail.html.haml
+++ b/app/views/research/submission_test_run_results/_fail.html.haml
@@ -1,4 +1,4 @@
-.test-run-result.fail
+.test-run-result.fail.open
   .header
     =icon(:"chevron-right", "Click to expand", style: :solid, extra_classes: 'closed-icon js-collapse-trigger')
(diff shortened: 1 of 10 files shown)


app/views/research/submission_test_run_results/_fail.html.haml
```diff
@@ -1,7 +1,7 @@
-.test-run-result.open.fail
+.test-run-result.fail
```
iHiD: Why has this changed? And the one in `_pass.html.haml`?

kntsoriano: The JS collapses everything by default. Did this use to work in a way that it was open by default?

iHiD: Yeah. Failed should always be open by default, and passed should always be collapsed by default. So this works to do that by having the different classes in the different files.

kntsoriano: Okay.

PR #907, opened 2020-02-06
Fix sneaky typo on supporters page

TheLostLambda:
Under Mozilla here: https://exercism.io/supporters


ErikSchierboom:
Thanks for catching this!

PR #915, opened 2020-02-26
Highlight test runs

kntsoriano:
Closes https://github.com/exercism/research_experiment_1/issues/86.

I decided to just highlight the code run and the expected values. Test output and test failure messages weren't highlighted. Let me know if you want that changed.

### DIFF
--- a/db/migrate/20200225110803_rename_syntax_highlighter_language.rb
+++ b/db/migrate/20200225110803_rename_syntax_highlighter_language.rb
@@ -1,0 +1,7 @@
+class RenameSyntaxHighlighterLanguage < ActiveRecord::Migration[6.0]
+  def change
+    add_column :tracks, :syntax_highlighter_language, :string
+
+    Track.update_all("syntax_highlighter_language = syntax_highligher_language")
+  end
+end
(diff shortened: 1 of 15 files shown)


db/migrate/20200225110803_rename_syntax_highlighter_language.rb
```diff
@@ -0,0 +1,7 @@
+class RenameSyntaxHighlighterLanguage < ActiveRecord::Migration[6.0]
+  def change
+    add_column :tracks, :syntax_highlighter_language, :string
+
+    Track.update_all("syntax_highlighter_language = syntax_highligher_language")
```
iHiD: Nice. This is a good way to do it, rather than renaming the column :+1:

db/schema.rb
```diff
@@ -637,6 +637,7 @@
     t.datetime "updated_at", null: false
     t.boolean "active", default: true, null: false
     t.integer "median_wait_time"
+    t.string "syntax_highlighter_language"
```
iHiD: The other one was `null: false`, so I think this should be too.

kntsoriano:
@iHiD Good catch! I've added the addition of the null constraint on a different migration as we first need to populate the field values and then apply the null constraint afterwards.

iHiD:
Great. Nice work. Thanks dude :)

PR #917, opened 2020-02-26
Add message when user doesn't have a browser with websockets

kntsoriano:
Closes https://github.com/exercism/internal-wip/issues/22.

As previously discussed, as a temporary solution, we just inform the user that they need to use websockets when doing the research experiment. I think we should change the copy on this modal to be better, but can't think of anything right now.

## Screenshots
Screen Shot 2020-02-26 at 4 41 34 PM

### DIFF
--- a/app/javascript/src/experiment-solution.js
+++ b/app/javascript/src/experiment-solution.js
@@ -6,6 +6,7 @@ import InfoPanel from './info_panel';
 import SubmitButton from './submit_button';
 import Submission from './submission';
 import PageThemeSelect from './page_theme_select';
+import WebsocketsCompatibilityMessage from './websockets_compatibility_message';
 
 class ExperimentSolution {
   constructor(element) {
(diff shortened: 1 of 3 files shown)


app/javascript/src/experiment-solution.js
```diff
@@ -26,6 +28,13 @@ class ExperimentSolution {
     this._setupSplit();
   }
 
+  _checkWebsocketsCompatibility() {
+    if(!('WebSocket' in window) && !('MozWebSocket' in window)) {
```
kntsoriano: I got this snippet from https://stackoverflow.com/questions/20060320/what-is-the-best-way-to-detect-websocket-support-using-javascript.

iHiD:
@ErikSchierboom Do things still work ok in the latest MS browser?

ErikSchierboom:
Yes, I've tested on IE 11 and that worked fine. I didn't see a popup :)

iHiD:
Great. Thanks :)

PR #918, opened 2020-02-26
Update Prism version

kntsoriano:
Closes https://github.com/exercism/wip/issues/62.

The CDN has just been updated :)

### DIFF
--- a/config/initializers/prism.rb
+++ b/config/initializers/prism.rb
@@ -1,7 +1,7 @@
 require "prism_js"
 
 PrismJS.configure do |config|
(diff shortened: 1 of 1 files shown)


kntsoriano:
Unfortunately, the issue isn't still fixed because Prism's 1.19 release doesn't include the fix for the issue. I'll merge this now as other issues might be fixed by this upgrade, but we can consider relying on Prism's master branch rather than their releases since there is a long interval between them.

kntsoriano:
I've deployed and syntax highlighting still works fine. We're now on the latest version of Prism on productionn.

iHiD:
Great. Thanks dude!

PR #934, opened 2020-04-22
Update octokit to fix deprecation errors

kntsoriano:
Relevant issue: https://github.com/octokit/octokit.rb/issues/1191

This issue is completely separate to the register issue we have on the Slack channel as this gem only works with pulling data from GitHub. We have a different gem for OAuth authentication for GitHub.


iHiD:
Thanks. Any ideas on the other issue?

PR #949, opened 2020-06-12
Return error in tracks GitApi when using invalid slug

ErikSchierboom:
For consistency and to prevent spamming GitHub with useless requests in came of a bug

### DIFF
--- a/app/controllers/git_api/tracks_controller.rb
+++ b/app/controllers/git_api/tracks_controller.rb
@@ -5,11 +5,17 @@ module GitAPI
     layout false
 
     def creation_issues
+      track = Track.find_by_slug(params[:id])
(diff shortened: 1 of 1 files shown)


app/controllers/git_api/tracks_controller.rb
```diff
@@ -63,5 +65,17 @@ def query
         }
       }
     end
+
+    def set_track
```
iHiD, line 68: This is quite important in Rails. because controller's methods can be exposed via the router, having things as `private` guarantees that a user won't be able to call this method directly.

```suggestion
    private
    def set_track
```

ErikSchierboom: Doh

PR #952, opened 2020-06-17
Conditionally write introduction.md file for new concept exercise

ErikSchierboom:
This is required as part of https://github.com/exercism/v3-dashboard/issues/18

### DIFF
--- a/app/controllers/git_api/concept_exercises_controller.rb
+++ b/app/controllers/git_api/concept_exercises_controller.rb
@@ -1,7 +1,7 @@
 module GitAPI
   class ConceptExercisesController < BaseController
     def create
-      # Validate presence and size of fields
+      # Validate presence and size of required fields
       %i{ 
         track_slug exercise_slug example_filename example_code
         design_markdown instructions_markdown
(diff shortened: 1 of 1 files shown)


app/controllers/git_api/concept_exercises_controller.rb
```diff
@@ -10,6 +10,9 @@ def create
         return render_error(:"too_large_#{param}") if params[param].size > 1.megabyte
       end
 
+      # Validate size of optional fields
+      return render_error(:"too_large_introduction_markdown") if params[:introduction_markdown].present? && params[:introduction_markdown].size > 1.megabyte
```
iHiD, line 14:
```suggestion
      return render_error(:too_large_introduction_markdown) if params[:introduction_markdown].present? && params[:introduction_markdown].size > 1.megabyte
```

iHiD: You could also use either `try` or `&.` with a `to_i` if you preferered. Just makes the line a bit more readable, but I don't mind either way.

```suggestion
      return render_error(:"too_large_introduction_markdown") if params[:introduction_markdown]&.size.to_i > 1.megabyte
```

PR #953, opened 2020-06-17
Use a human-readable branch name for new concept-exercise PRs

ErikSchierboom:
Closes https://github.com/exercism/v3-dashboard/issues/52

### DIFF
--- a/app/controllers/git_api/concept_exercises_controller.rb
+++ b/app/controllers/git_api/concept_exercises_controller.rb
@@ -28,21 +28,19 @@ module GitAPI
         `git remote add upstream https://github.com/exercism/v3.git`
         `git fetch upstream master`
         `git checkout --no-track -b #{branch_name} upstream/master`
-      end
 
-      exercise_path = tmp_path / "languages" / params[:track_slug] / "exercises" / "concept" / params[:exercise_slug]
-      exercise_docs_path = exercise_path / ".docs"
-      exercise_meta_path = exercise_path / ".meta"
+        exercise_path = tmp_path / "languages" / params[:track_slug] / "exercises" / "concept" / params[:exercise_slug]
+        exercise_docs_path = exercise_path / ".docs"
+        exercise_meta_path = exercise_path / ".meta"
 
-      Dir.mkdir(exercise_path)
-      Dir.mkdir(exercise_docs_path)
-      Dir.mkdir(exercise_meta_path)
+        Dir.mkdir(exercise_path)
+        Dir.mkdir(exercise_docs_path)
+        Dir.mkdir(exercise_meta_path)
 
-      File.open(exercise_docs_path / "instructions.md", "w") {|f|f.write params[:instructions_markdown] }
-      File.open(exercise_meta_path / "design.md", "w") {|f|f.write params[:design_markdown] }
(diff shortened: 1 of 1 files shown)


app/controllers/git_api/concept_exercises_controller.rb
```diff
@@ -20,29 +20,26 @@ def create
...
       uuid = SecureRandom.uuid.gsub('-', '')
       tmp_path = Pathname.new("/tmp/#{uuid}")
-      branch_name = "#{params[:track_slug]}/#{params[:exercise_slug]}-#{uuid}"
 
       `git clone https://#{Rails.application.secrets.exercism_bot_credentials}@github.com/exercism-bot/v3.git #{tmp_path}`
 
       Dir.chdir(tmp_path) do
         `git remote add upstream https://github.com/exercism/v3.git`
-        `git fetch upstream master`
+        `git fetch upstream`
```
ErikSchierboom, line 28: I have to retrieve all branch names now to detect if the branch already exists

app/controllers/git_api/concept_exercises_controller.rb
```diff
@@ -20,29 +20,26 @@ def create
...
+        Dir.mkdir(exercise_path)
+        Dir.mkdir(exercise_docs_path)
+        Dir.mkdir(exercise_meta_path)
 
-      File.open(exercise_docs_path / "instructions.md", "w") {|f|f.write params[:instructions_markdown] }
-      File.open(exercise_meta_path / "design.md", "w") {|f|f.write params[:design_markdown] }
-      File.open(exercise_path / params[:example_filename], "w") {|f|f.write params[:example_code] }
+        File.open(exercise_docs_path / "instructions.md", "w") {|f|f.write params[:instructions_markdown] }
+        File.open(exercise_meta_path / "design.md", "w") {|f|f.write params[:design_markdown] }
+        File.open(exercise_path / params[:example_filename], "w") {|f|f.write params[:example_code] }
```
ErikSchierboom, line 41: I've moved all code in a single `Dir.chdir` block to make it more clear that they are all working on the same directory

app/controllers/git_api/concept_exercises_controller.rb
```diff
@@ -61,5 +58,26 @@ def render_error(type)
         error: { type: type }
       }, status: 400
     end
+
+    private
+
+    def branch_name
+      @branch_name ||= begin
```
ErikSchierboom, line 65: I've used memoization here to only calculate the branch name once (it is used twice).

app/controllers/git_api/concept_exercises_controller.rb
```diff
@@ -61,5 +58,26 @@ def render_error(type)
...
+          suffix += 1
+        end
+
+        potential_branch_name
+      end
+    end
+
+    def branch_name_not_unique(branch_name)
+      `git show-ref --verify --quiet refs/remotes/origin/#{branch_name}`
+      return $?.success?
```
iHiD, line 80: You can use [system](https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-system) which returns true if the command is a success and false otherwise.

```suggestion
      system("git show-ref --verify --quiet refs/remotes/origin/#{branch_name}")
```

PR #969, opened 2020-09-30
Update README.md

printfMayank:


kotp:
The sentiment is appreciated, but not sure that it should be a new heading level.

Might save this change for when there is a bit of a larger reorganization that needs to be done?

PR #999, opened 2021-06-02
Small grammatical change

eikooc:
Fixed the classic you're vs. your


iHiD:
@eikooc Thanks for this. The existing website is now redundant (we rewrote it from scratch) pending the launch of Exercism v3, and no longer deployed. So this won't make it - sorry. But thank you!

